PR #1, opened 2022-05-25
Implement requires side relation charm library

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-413](https://warthogs.atlassian.net/browse/DPE-413).

In long:
* Application charms need an easy way to communicate through relations to the database charms from Canonical Data Platform.
* This communication should be handled in a common way for whichever database charm the application is communicating to.
* The library should notify the application charm about what the database charm changed in the relation databag every time it triggers a relation changed event.

# Solution
* Created a charm library exposing the fields listed in the specification from https://github.com/canonical/charm-relation-interfaces/pull/2.
* To notify the application charm about what changed in the relation databag when a relation changed event happens a diff mechanism was implemented using another field in the relation databag to store a snapshot of the previous state of the fields in the relation databag (the state that the charm had in the previous relation changed event that happened).
* After the field that has changed is identified, a custom event is emitted to the application charm, that should observe it and handle it as it normally handle the default charms events.
* Peers relations were not used to store the previous state used by the diff because they would enforce the application charm to create a peer relation, even when not desired to have one.
* Stored state was not used because there are some [concerns](https://juju.is/docs/sdk/stored-state-uses-limitations) about it, specially in k8s charms (when there is the need to use the flag `use_juju_for_storage`, which is not recommended).
* Writing the previous state to files in the VM or container was not considered because maybe it can cause some issue because of a conflict with what the application charm is writing to the filesystem.

# Context
* The provides side charm library is implemented in https://github.com/canonical/data-platform-libs/pull/2. It can be reviewed and merged in parallel.
* **The files that you should focus more when reviewing are the following** (as the list of new files is big):
    * `.gitignore`
    * `CONTRIBUTING.md`
    * `README.md`
    * `lib/charms/data_platform_libs/v0/database_requires.py`
    * `tests/unit/test_database_requires.py`
* The other files are mostly files imported from the operator template and a placeholder charm, as this is a libraries' repository (but they should also be reviewed).
* The number of custom events that are defined is small to not confuse application charm developers. They are the following:
    * `database_created`: the database charm created the database with the name provided by the application charm and has shared the username and the password (available as properties in the library).
    * `endpoints_changed`: the database charm changed its read/write endpoints (they're available as a property in the library).
    * `read_only_endpoints_changed`: the database charm changed its read-only endpoints (they're also available as a property in the library).
* Some additional fields are also exposed in the library (also following the specification mentioned earlier):
    * `repset`: MongoDB replica set name.
    * `tls`: whether TLS is enables in the database connection.
    * `tls-ca`: TLS CA.
    * `uris`: URIs for Kafka, MongoDB, OpenSearch and Redis.
    * `version`: version of the database deployed by the charm.
* And also some methods are exposed in the library to enable the application charm to interact with the database charm through the relation:
    * `set_database`: to request a database with a specific name.
    * `set_extra_user_roles`: to request extra roles for the user that was created for the application (it's useful for proxy application like PgBouncer, which needs permission to create new databases and users).

# Testing
* Unit tests were implemented for the requires side charm library.
* Integration tests are implemented on https://github.com/canonical/data-platform-libs/pull/3 (which should be merged only after this PR and https://github.com/canonical/data-platform-libs/pull/2 are merged).

# Release Notes
* Add requires side charm library.

# Thanks
* Thanks to Mykola (specification), Paulo (ideas and documentation), Shayan and Will (ideas), Pen and Pietro (ideas about the diff).


requirements.txt
```diff
@@ -0,0 +1 @@
+ops >= 1.2.0
```
paulomach, line 1: should we bump it to 1.4?

paulomach (commented):
LGTM, but I'm biased on this, so i'll keep the approvals for others

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,267 @@
...
+from charms.data_platform_libs.v0.database_requires import DatabaseRequires
+
+class ApplicationCharm(CharmBase):
+    # Application charm that connects to database charms.
+
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        # Charm events defined in the database requires charm library.
+        self.database = DatabaseRequires(self, "database")
```
jnsgruk, line 35: Just to be super explicit for documentation's sake, I'd specify this as a keyword arg:

```suggestion
        self.database = DatabaseRequires(self, relation_name="database")
```

jnsgruk (approved):
This is a nice start, thanks Marcelo

PietroPasotti:
Many of the files added by charmcraft init are not needed, and I think they should be removed.
Take a look at https://github.com/canonical/charm-lib-template
I believe the only files that are 'required' to create and publish a lib are metadata.yaml (only the 'name' field is necessary in there) and /lib/charms/.../lib_file.py

So if you want to de-clutter the PR, /src should go. as well as .jujuignore and charmcraft.yaml

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,267 @@
...
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+
+class DatabaseCreatedEvent(RelationEvent):
+    """Event emitted when a new database is created for use on this relation."""
+
+    pass
```
PietroPasotti, line 79: pass is not necessary since there's a docstring

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Args:
+            event: relation changed event.
+
+        Returns:
+            a dict containing the added, deleted and changed
+                keys from the event relation databag.
+        """
+        # Retrieve the old data from the data key in the application relation databag.
+        old_data = json.loads(self.relation.data[self.charm.model.app].get("data", "{}"))
```
PietroPasotti, line 130: As mentioned in a call earlier today: are we sure that the application databag can be safely used to cache data? I.e. if the app is scaled to 0 and back up to N, will the application data be wiped or persisted?
If the remote application does?

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,267 @@
...
+    @property
+    def uris(self) -> str:
+        """Returns the connection URIs.
+
+        MongoDB, Redis, OpenSearch and Kafka only.
+        """
+        return self._get_relation_data("uris")
+
+    @property
+    def username(self) -> str:
```
PietroPasotti, line 214: the return type of this prop (and many others) should be Optional[str]

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        # TODO: evaluate the possibility of losing the diff if some error
+        # happens in the charm before the diff is completely checked (DPE-412).
+        # Convert the new_data to a serializable format and save it for a next diff check.
+        data = {key: value for key, value in new_data.items() if key != "data"}
+        self._update_relation_data("data", json.dumps(data))
+
+        # Return the diff with all possible changes.
+        return {
```
PietroPasotti, line 151: that's just me, but I'd create a `Diff` namedtuple to type this return object and make its consumption easier (and more maintainable) etc... But consider this a nit :D

PietroPasotti (changes requested):
Some minor comments, some of which need double checking :)

lib/charms/data_platform_libs/v0/database_requires.py
PietroPasotti: (that applies to the other events too)

PietroPasotti:
Looking a lot better now! Green light from my side :)

WRFitch (approved):

requirements.txt
marceloneppel: Updated on https://github.com/canonical/data-platform-libs/pull/1/commits/8e7f1083d4b3621e85f6aa77152285833f423e39.

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: Nice! Updated on https://github.com/canonical/data-platform-libs/pull/1/commits/ba58e91a5c7b2822d84e60089c318f25774c6ec7.

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: Fixed on https://github.com/canonical/data-platform-libs/pull/1/commits/3ab2626f83db01fc89a25c9c6e13558d3e9d62e1.

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: John Meinel confirmed to me yesterday that it should work: "Scaling to 0 units is not removing the application. Thus I would expect the unit data to go away, but the application data in the relation to still be there."

tox.ini
```diff
@@ -0,0 +1,75 @@
...
+    flake8-docstrings
+    flake8-copyright
+    flake8-builtins
+    pyproject-flake8
+    pep8-naming
+    isort
+    codespell
+commands =
+    # uncomment the following line if this charm owns a lib
+    # codespell {[vars]lib_path}
```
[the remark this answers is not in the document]
marceloneppel: Nice. Thanks for catching it. Fixed on https://github.com/canonical/data-platform-libs/pull/1/commits/39c1e7388ea6ba4df0b0fce519b6863a373763d1.

PR #4, opened 2022-06-08
Handle multiple applications connecting to the database

marceloneppel:
# Issue
In short this PR address JIRA ticket [DPE-418](https://warthogs.atlassian.net/browse/DPE-418).

In long:
* Database charms using the provides charm library should be able to handle multiple applications requesting new databases.
* Currently, the library can't handle this situation. It overrides the data from the first application in the relation databag because it's using the same relation without providing a relation id (the id that differentiates the connections being made to the same relation in the database charm).

# Solution
* With the relation id it s possible to know which application is requesting the database and also send the credentials back.

# Context
* The documentation in the top of the library and also the database test charm were updated with the correct way to handle the relation data from one or multiple applications.
* The properties from the `DatabaseProvides` class were moved to the `DatabaseEvent` class, which is a superclass for the custom events of this library.
* The functions from the `DatabaseProvides` class now need the relation id to be passed as a parameter.
* In the `DatabaseProvides` class, there is a new function called `fetch_relation_data` to make it still be possible to retrieve the data from the relation when outside an event callback like `_on_database_requested`.

# Testing
* Unit tests were updated and some new tests (related to setting data using the event object, which is now used to set data in the right relation instance) were implemented.
* One more integration test was added to cover the case of two applications requesting new databases to the same database charm.

# Release Notes
* Make it possible to handle multiple applications requesting new databases in the provides charm library.

### DIFF
--- a/lib/charms/data_platform_libs/v0/database_provides.py
+++ b/lib/charms/data_platform_libs/v0/database_provides.py
@@ -219,9 +219,6 @@ class DatabaseProvides(Object):
         This function writes in the application data bag, therefore,
         only the leader unit can call it.
 
-        This function can be used to set the data
-        when outside an event callback.
-
         Args:
             relation_id: the identifier for a particular relation.
             username: user that was created.
@@ -241,9 +238,6 @@ class DatabaseProvides(Object):
         This function writes in the application data bag, therefore,
         only the leader unit can call it.
 
-        This function can be used to set the data
-        when outside an event callback.
-
         Args:
             relation_id: the identifier for a particular relation.
             connection_strings: database hosts and ports comma separated list.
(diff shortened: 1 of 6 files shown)


lib/charms/data_platform_libs/v0/database_provides.py
```diff
@@ -75,7 +75,148 @@ def _on_database_requested(self, _) -> None:
...
+        """Retrieves data from relation.
+
+        Args:
+            key: key to retrieve the data from the relation.
+
+        Returns:
+            value stored in the relation data bag for
+                the specified key or None if the key doesn't exist.
+        """
+        return self.relation.data[self.relation.app].get(key)
```
jnsgruk, line 111: Some of these methods feel like they don't belong as properties on the event itself? I'm not sure I understand why there are methods on this event for mutating properties? The event should really be used to just carry information...

delgod (approved):

lib/charms/data_platform_libs/v0/database_provides.py
jnsgruk: Nice, thanks!

jnsgruk (approved):

paulomach (approved):
LGTM

PR #5, opened 2022-06-08
Add partial code to handle multiple clusters in the same relation

marceloneppel:
# Issue
In short this PR partially address JIRA ticket [DPE-423](https://warthogs.atlassian.net/browse/DPE-423).

In long:
* Application charms using the requires charm library should be able to connect to multiple database clusters through the same relation.
* Currently, the library can't handle this situation. It faces some errors if we try to connect multiple clusters to the same relation because it doesn't provide a relation id (the id that differentiates the connections being made to the same relation in the database charm) when interacting with each different database cluster.

# Solution
* With the relation id it's possible to know which database cluster the application is talking to at a specific moment and get the relation data specific to that cluster.

# Context
* The documentation in the top of the library and also the database test charm were updated with the correct way to handle the relation data from one or multiple applications.
* The properties from the `DatabaseRequires` class  were moved to the `DatabaseEvent` class, which is a superclass for the custom events of this library. This is responsible for the big diff in the `DatabaseEvent` class.
* The functions from the `DatabaseRequires` class now use the relation id to correctly emit events from each different cluster.
* In the `DatabaseRequires` class, there is a new function called `fetch_relation_data` to make it still be possible to retrieve the data from the relation when outside an event callback like `_on_database_created`.
* **In summary, this PR addresses part of the implementation needed for the library to be able to connect to multiple database clusters. The remaining implementation will be done in the next PR (to avoid a bigger PR).**

# Testing
* Unit tests were updated and some new tests (related to setting data using the event object, which is now used to set data in the right relation instance) were implemented.
* Integration tests were not updated yet as there is some more code that needs to be implemented in the next PR.

# Release Notes
* Add code needed for later implementation about connecting to multiple database clusters through the same relation using the requires charm library.

### DIFF
--- a/lib/charms/data_platform_libs/v0/database_requires.py
+++ b/lib/charms/data_platform_libs/v0/database_requires.py
@@ -35,14 +35,14 @@ class ApplicationCharm(CharmBase):
         self.database = DatabaseRequires(self, relation_name="database", database_name="database")
         self.framework.observe(self.database.on.database_created, self._on_database_created)
 
-    def _on_database_created(self, _) -> None:
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
         # Handle the created database
 
         # Create configuration file for app
         config_file = self._render_app_config_file(
-            self.database.username,
-            self.database.password,
-            self.database.endpoints,
+            event.username,
+            event.password,
+            event.endpoints,
         )
 
         # Start application with rendered configuration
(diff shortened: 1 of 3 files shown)


delgod (approved):

jnsgruk (approved):
Assuming integration tests to follow, LGTM. Thanks :)

PR #6, opened 2022-06-08
Update tester charm and add test for multiple databases requests

marceloneppel:
# Issue
In short this PR address JIRA ticket [DPE-422](https://warthogs.atlassian.net/browse/DPE-422).

In long:
* Database charms using the requires charm library should be able to request multiple databases to the same database charm.
* Currently, this is something that isn't being tested.

# Solution
* Add an integration test that ensures that the requires charm library can request multiple databases to the same database charm.

# Context
* Multiple relations are needed to accomplish that, because we can't relate twice the same application and database using the same relation name.
* Some updates were made in the `tests/integration/helpers.py`and `tests/integration/test_charm.py` files to handle .

# Testing
* Unit tests were not updated on this PR as there wasn't any update in the libraries code.
* One more integration test was added to cover the case of two databases being requested by the application to the same database charm.
* Some updates were made in the integration tests to use some constants related to the relation names.

# Release Notes
* Ensure that is possible to request multiple databases to the same database charm using the requires charm library.

### DIFF
--- a/tests/integration/application-charm/metadata.yaml
+++ b/tests/integration/application-charm/metadata.yaml
@@ -8,7 +8,5 @@ summary: |
   only for testing of the libs in this repository.
 
 requires:
-  first-database:
-    interface: database-client
-  second-database:
+  database:
     interface: database-client
--- a/tests/integration/application-charm/metadata.yaml
+++ b/tests/integration/application-charm/metadata.yaml
@@ -8,5 +8,7 @@ summary: |
   only for testing of the libs in this repository.
 
 requires:
-  database:
+  first-database:
+    interface: database-client
+  second-database:
     interface: database-client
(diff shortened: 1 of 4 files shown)


tests/integration/application-charm/metadata.yaml
```diff
@@ -8,5 +8,7 @@ summary: |
   only for testing of the libs in this repository.
 
 requires:
-  database:
+  first-database:
+    interface: database-client
+  second-database:
```
shayancanonical, line 13: how will this work with N databases? is it possible to set up a way to transfer database names/credentials/other info through the same relation? the legacy mysqlrouter charm does this, so im happy to provide a reference for what i was thinking of

delgod (approved):

MiaAltieri (approved):
LGTM! Looks great Marcelo!

tests/integration/application-charm/metadata.yaml
shayancanonical: Nevermind, I think I finally understand what is going on

PR #7, opened 2022-06-09
Finish implementation of connection to multiple cluster with relation aliases

marceloneppel:
# Issue
In short this PR address JIRA ticket [DPE-423](https://warthogs.atlassian.net/browse/DPE-423).

In long:
* Application charms using the requires charm library should be able to connect to multiple database clusters through the same relation.
* Currently, the library can handle it, but there is no example or test showing how to do that.
* The custom events names are hardcoded in the library, similar to what is described on https://github.com/canonical/redis-k8s-operator/issues/30. 

# Solution
* The items below were previously implemented in https://github.com/canonical/data-platform-libs/pull/5:
  * With the relation id it's possible to know which database cluster the application is talking to at a specific moment and get the relation data specific to that cluster.
* Add an option to assign an alias to each relation/cluster and use it in the custom events names, so an application charm developer can handle the events in both ways (using just one event observer for multiple clusters or different observers for each cluster).

# Context
* No more code were added to the library related to the change to use the relation id. Everything was functional in the last PR.
* This PR focused on adding the missing integration test related to the connection to multiple clusters through the same relation.
* It also focused on adding the option to add an alias to each relation/cluster and use it in the custom events names (so the event names aren't simply hardcoded - there are still the hardcoded events, but now the application charm developer can opt to use different events for different clusters). For that, some code was added to the `DatabaseRequires` class:
  * `relations_aliases` attribute that store the relations/clusters aliases and is used to create the custom events names for each relation/cluster.
  * `_assign_relation_alias` method: used to assign an alias to a relation instance after `juju relate` is called.
  * `_emit_custom_event` method: used to emit a custom event for a specific relation/cluster.
  * `_get_relation_alias` method: used to get the alias of a relation instance.
* The application tester charm was updated to have an example of how to use the library when it's desired to connect to multiple clusters through the same relation and using aliases for each relation.

# Testing
* Unit tests were updated and some new tests (related to the alias feature) were implemented.
* One more integration test was implemented to check that an application using the requires side charm library can connect to multiple clusters using the same relation.

# Release Notes
* Finish implementation of connection to multiple clusters using the same relation.

### DIFF
--- a/lib/charms/data_platform_libs/v0/database_requires.py
+++ b/lib/charms/data_platform_libs/v0/database_requires.py
@@ -35,14 +35,14 @@ class ApplicationCharm(CharmBase):
         self.database = DatabaseRequires(self, relation_name="database", database_name="database")
         self.framework.observe(self.database.on.database_created, self._on_database_created)
 
-    def _on_database_created(self, _) -> None:
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
         # Handle the created database
 
         # Create configuration file for app
         config_file = self._render_app_config_file(
-            self.database.username,
-            self.database.password,
-            self.database.endpoints,
+            event.username,
+            event.password,
+            event.endpoints,
         )
 
         # Start application with rendered configuration
@@ -57,10 +57,16 @@ import json
 import logging
 from collections import namedtuple
 from datetime import datetime
-from typing import Optional
+from typing import List, Optional
 
-from ops.charm import CharmEvents, RelationChangedEvent, RelationEvent
+from ops.charm import (
+    CharmEvents,
+    RelationChangedEvent,
+    RelationEvent,
+    RelationJoinedEvent,
+)
 from ops.framework import EventSource, Object
+from ops.model import BlockedStatus, Relation
 
 # The unique Charmhub library identifier, never change it
 LIBID = "0241e088ffa9440fb4e3126349b2fb62"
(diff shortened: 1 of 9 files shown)


lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -267,6 +347,15 @@ def _update_relation_data(self, relation_id: int, data: dict) -> None:
...
     def _on_relation_joined_event(self, event: RelationJoinedEvent) -> None:
         """Event emitted when the application joins the database relation."""
+        # If relations aliases were provided, assign one to the relation.
+        try:
+            self._assign_relation_alias(event.relation.id)
+        except NoAliasAvailableError as e:
+            logger.error(str(e))
+            # Set a blocked status if there is no available alias
+            # (all of them were already assigned to other relations).
+            self.charm.unit.status = BlockedStatus(str(e))
```
paulomach, line 357: I don't like the idea of set unit status within the charm library.
I'm afraid this can get too hidden and hard to pin when things go wrong. 
(I actually advocate to only do status setting in the charm code.)

Anyway, my suggestion not treat the exception here.

paulomach (commented):
I left one blocking (to me) comment. Want to see more comments on that.
All the rest and tests looks good though.

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -191,13 +196,63 @@ def __init__(
...
+        if not self.local_unit.is_leader():
+            return
+
+        # If no aliases were provided, return immediately.
+        if not self.relations_aliases:
+            return
+
+        # Retrieve the available aliases (the ones that weren't assigned to any relation).
+        available_aliases = self.relations_aliases[:]
+        print(available_aliases)
```
WRFitch, line 241: Should this be logged somewhere using `logger.info`, rather than `print`?

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -191,13 +196,63 @@ def __init__(
...
+        for relation in self.charm.model.relations[self.relation_name]:
+            alias = relation.data[self.local_app].get("alias")
+            if alias:
+                print(f"Alias {alias} was already assigned to relation {relation.id}")
```
WRFitch, line 245: see above. If I'm missing a valid reason to use `print`, fair enough, but otherwise could these be changed to logger.info?

tests/integration/test_charm.py
```diff
@@ -118,12 +128,61 @@ async def test_two_applications_doesnt_share_the_same_relation_data(
...
+    # and wait for them exchanging some connection data.
+    await ops_test.model.add_relation(
+        f"{APPLICATION_APP_NAME}:{MULTIPLE_DATABASE_CLUSTERS_RELATION_NAME}", DATABASE_APP_NAME
+    )
+    # await ops_test.model.wait_for_idle(apps=all_app_names, status="active")
+
+    await ops_test.model.add_relation(
+        f"{APPLICATION_APP_NAME}:{MULTIPLE_DATABASE_CLUSTERS_RELATION_NAME}",
+        another_database_app_name,
+    )
```
WRFitch, line 177: You could join a few of these with an `asyncio.gather()` if you don't want to test them individually. Up to you, though.

WRFitch (commented):
LGTM, a couple of minor nits about style but this looks really solid :)

delgod (approved):

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: No. It shouldn't be logged. I removed it (https://github.com/canonical/data-platform-libs/pull/7/commits/119d1c8494cc5518360d4813c0241788607771c8). Thanks for catching it.

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: Changed to `logger.debug` (https://github.com/canonical/data-platform-libs/commit/119d1c8494cc5518360d4813c0241788607771c8).

tests/integration/test_charm.py
marceloneppel: Good suggestion. Implemented on https://github.com/canonical/data-platform-libs/pull/7/commits/0858e5c739d545f5cfbebdbf53f73ba8b408c636.

lib/charms/data_platform_libs/v0/database_requires.py
marceloneppel: I agree with this point. I checked with the operator framework team and figured out that there is no way to get the error in the charm code.

Another option would be to move the `on relation joined` handling to the charm code, but it would force every application charm using this library to write the data to the relation instead of relying on this library.

So, I found another option, that seems better (also to avoid errors in the charm; the error will just happen during deployment if the aliases are set up incorrectly).

This option checks if the number of relation aliases is equal to the number of maximum number of connections (`limit` field) that are informed in the relation section on `metadata.yaml`. This way, the list of cluster names will always match the number of possible `juju relate` commands that the operator can call.

One last detail: if the application charm is not using the cluster aliases, it's not required to inform the limit field in `metadata.yaml`.

marceloneppel:
@delgod, I implemented a test with the multiple database clusters withouth alias (https://github.com/canonical/data-platform-libs/pull/7/commits/47995f16333b3df0714add28329bbefb688ae33f).

lib/charms/data_platform_libs/v0/database_requires.py
paulomach: Agree about the documentation needs

paulomach (approved):

PR #8, opened 2022-06-16
Improve documementation

marceloneppel:
# Issue
In short this PR address JIRA ticket [DPE-460](https://warthogs.atlassian.net/browse/DPE-460).

In long:
* The current requires charm library doesn't document how to use cluster aliases and how to differentiate multiple clusters in the library header.
* Both libraries doesn't explain the available custom events in the header documentation.
* Also, sometime the provides charm library generates some confusion when a database charm already has a provides library.

# Solution
* Add some more details and examples (in the case of the cluster aliases) in the libraries header documentation.

# Context
* For the third item mentioned in the issue section, MongoDB, for example, has its own [provides charm library](https://github.com/canonical/mongodb-operator/blob/main/lib/charms/mongodb_libs/v0/mongodb_provider.py), which also handles the creation of database and users.

# Testing
* No new tests are needed as what has changed was only documentation.

# Release Notes
* Add documentation about connecting to multiple clusters
* Add documentation about custom events
* Document the possibility about using the provides charm library as a template for more complete libraries

### DIFF
--- a/lib/charms/data_platform_libs/v0/database_requires.py
+++ b/lib/charms/data_platform_libs/v0/database_requires.py
@@ -52,6 +52,14 @@ class ApplicationCharm(CharmBase):
         self.unit.status = ActiveStatus("received database credentials")
 ```
 
+As shown above, the library provides some custom events to handle specific situations,
+which are listed below:
+
+- database_created: event emitted when the requested database was created
+- endpoints_changed: event emitted when the read/write endpoints of the database have changed
+- read_only_endpoints_changed: event emitted when the read-only endpoints of the database
+  have changed
+
 If it's needed to connect multiple database clusters to the same relation endpoint
 it's possible to provide or not aliases for the different clusters/relations.
 
--- a/lib/charms/data_platform_libs/v0/database_requires.py
+++ b/lib/charms/data_platform_libs/v0/database_requires.py
@@ -51,6 +51,70 @@ class ApplicationCharm(CharmBase):
         # Set active status
         self.unit.status = ActiveStatus("received database credentials")
 ```
+
+If it's needed to connect multiple database clusters to the same relation endpoint
+it's possible to provide or not aliases for the different clusters/relations.
+
+To differentiate multiple clusters connected to the same relation endpoint
+without providing aliases, it's possible to use the name of the remote application:
+
+```python
+
+def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+    # Get the remote app name of the cluster that triggered this event
+    cluster = event.relation.app.name
+```
+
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -51,6 +51,78 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+
+```python
+
+def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+    # Get the remote app name of the cluster that triggered this event
+    cluster = event.relation.app.name
+```
+
+If cluster aliases are provided, then it's possible to differentiate the clusters in two ways.
+The first is the same as without providing cluster aliases (using the remote application name).
```
MiaAltieri, line 77:
```suggestion
The first is to use the remote application name, ie `event.relation.app.name`, as mentioned above.
```

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -51,6 +51,78 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+
+The second way to do that is by using different event handlers to handle each cluster events.
+The implementation would be something like the following code:
```
MiaAltieri, line 80:
```suggestion
The second way is to use different event handlers to handle each cluster events.
The implementation would be something like the following code:
```

MiaAltieri (approved):
This looks awesome! Great documention and super thorough, comments are just minor knits.

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -51,6 +51,78 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
+As shown above, the library provides some custom events to handle specific situations,
+which are listed below:
+
+- database_created: event emitted when the requested database was created
+- endpoints_changed: event emitted when the read/write endpoints of the database have changed
+- read_only_endpoints_changed: event emitted when the read-only endpoints of the database
+  have changed
+
+If it's needed to connect multiple database clusters to the same relation endpoint
+it's possible to provide or not aliases for the different clusters/relations.
```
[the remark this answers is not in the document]
marceloneppel: Thanks Mia! I changed it on https://github.com/canonical/data-platform-libs/pull/8/commits/2d9333f72bf090b68820c8e737bc3d95ab5326ad and also later on https://github.com/canonical/data-platform-libs/pull/8/commits/89444da9a7944d42bbc1615cf546d4a8cd5fd7c7.

zmraul (approved):
Great addition to the docs ✨

PR #11, opened 2022-06-23
add the ability to stop service when database relations gone

delgod:


jnsgruk (commented):
Change looks sane, but needs a test :)

delgod:
The "database_requires" and "database_provides" libraries are now deprecated in favour of "data_interfaces", which should provide a more comprehensive framework to extend to other technologies beside databases. On this migration, the signature of "DatabaseRequires" and "DatabaseProvides" did not change.
PR should be created for a new library.

PR #12, opened 2022-06-23
allow non-leader units receive database credentials

delgod:
# Issue
Non-leader units are not receiving the database events emitted by the requires library. This way they cannot know when the database credentials are received.

# Solution
Remove the checks about the unit being the leader and also move some data, like the diff (data field) and the alias fields to the unit databag (currently, they are part of the application databag).

# Testing
* Existing unit and integration tests were fixed.
* One of the existing unit tests (`test_database_events`) was updated to ensure that the events are emitted to both leader and non-leader units.

# Release Notes
* Allow non-leader units to receive library events and database credentials.


MiaAltieri (approved):

MiaAltieri:
to resolve failing lint see [this issue](https://github.com/issues)

tests/unit/test_database_requires.py
```diff
@@ -93,10 +94,12 @@ def test_diff(self):
         # Define a mock relation changed event to be used in the subsequent diff calls.
         mock_event = Mock()
         # Set the app, id and the initial data for the relation.
-        mock_event.app = self.harness.charm.model.get_app("application")
+        mock_event.app = self.harness.charm.model.get_app("database")
+        local_unit = self.harness.charm.model.get_unit("application/0")
```
WRFitch, line 98: you could set a variable for `"application/0"` if you wanted to minimise updates.

WRFitch (approved):

PR #13, opened 2022-06-24
avoid unnecessary application restarts.

delgod:
# Issue
If multiple relation data fields like, for example, username, password and endpoints are changed (even at the same time in the provides side/database side), the requires side libary will trigger multiple events at once  (like database created and endpoints changed).

This, in some application charms, can cause multiple restarts of the application, because some applications would need to restart in both situations 

# Solution
* Don't trigger endpoints changed event if a database created event was trigerred in the same relation changed event.
* Don't trigger read only endpoints changed event if an endpoints changed or a database created event was trigerred in the same relation changed event.

# Testing
* The existing unit test related to emitting events was updated to ensure that multiple event are not triggered.
* It was also manually tested the new unit test code with the previous library code (it correctly fails).

# Release Notes
* Change the requires library to avoid triggering multiple events at once


jnsgruk (commented):
Change looks sane, but surprised this hasn't resulted in a test change

marceloneppel:
Test implemented.

paulomach (approved):
lgtm

hey @PietroPasotti , a posterchild for `charms.harness_extensions.v0.capture_events`

tests/unit/test_database_requires.py
```diff
@@ -339,39 +339,79 @@ def test_database_events(self, is_leader: bool):
         # and the non-leader units through is_leader parameter.
         self.harness.set_leader(is_leader)
 
+        # Define the events that need to be emitted.
+        # The event key is the event that should have been emitted
+        # and the data key is the data that will be updated in the
+        # relation databag to trigger that event.
         events = [
```
MiaAltieri, line 346: maybe this can be stored in a different file and then read in here as to avoid this large variable, just a thought

tests/unit/test_database_requires.py
```diff
@@ -339,39 +339,79 @@ def test_database_events(self, is_leader: bool):
...
 
             # There are two events (one aliased and the other without alias).
             assert len(captured_events) == 2
+
+            # Check that the events that were emitted are the ones that were expected.
+            assert all(
+                isinstance(captured_event, event["event"]) for captured_event in captured_events
+            )
+
+            # Test that the remote unit name is available in the event.
```
MiaAltieri, line 415: excellent test

MiaAltieri (approved):

MiaAltieri (approved):

tests/unit/test_database_requires.py
marceloneppel: Yeah. It seems a good idea. The only info that needs to be here is the Classes references. I will keep them here for now.

Issue #14, opened 2022-07-04
DatabaseRequestedEvent dont set `app` or `unit`


paulomach:
### Bug Description

The emitter of `DatabaseRequestedEvent` don't set values to the `event.unit` nor `event.app` properties inherited from `RelationEvent`, so they are both defaulted to `None`.

For MySQL, the remote `unit.name` is used to index, within the database, the units specific user (e.g. `user`@`address`).

Also, without setting these properties, direct  access to relation remote app/unit databag is not possible, .e.g.:
```
(Pdb) event.relation.data[event.unit]
*** KeyError: None
```

### To Reproduce

1.Add a handler for `DatabaseRequestedEvent`
2.Inside the handler, try access `event.app` or `event.unit`

### Environment

Tested with v0.1 from charmhub

### Relevant log output

```shell
2022-07-04 21:14:26,327 ERROR    Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 281, in <module>
    main(MySQLOperatorCharm)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/main.py", line 431, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 316, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 784, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 857, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-mysql-81/charm/lib/charms/data_platform_libs/v0/database_provides.py", line 187, in _on_relation_changed
    self.on.database_requested.emit(event.relation)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 316, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 784, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-mysql-81/charm/venv/ops/framework.py", line 857, in _reemit
    custom_handler(event)
[...]

PR #15, opened 2022-07-05
Add app and unit information to events

marceloneppel:
# Issue
The emitter of DatabaseRequestedEvent don't set values to the event.unit nor event.app properties inherited from RelationEvent, so they are both defaulted to None https://github.com/canonical/data-platform-libs/issues/14.

The same applies to the requires side of the relation.

# Solution
* Use the app and the unit from the original event in the function that emits the custom events.

# Context
* Skip `lib/charms/harness_extensions/v0/capture_events.py` when reviewing this PR (it is an imported library). For more details, check [this](https://discourse.charmhub.io/t/harness-recipe-capture-events/6581).

# Testing
* Unit tests were updated and some new tests were implemented.

# Release Notes
* Add app and unit information to custom events.

### DIFF
--- a/lib/charms/data_platform_libs/v0/database_provides.py
+++ b/lib/charms/data_platform_libs/v0/database_provides.py
@@ -80,7 +80,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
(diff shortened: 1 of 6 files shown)


zmraul (approved):
Neat! I really like capturing events on unit tests. Taking notes for my own tests :)

delgod (approved):

PR #20, opened 2022-09-14
Include DatabaseCreatedEvent in sample code

mthaddon:
Trivial change updating the docstring for the database requires side of the relation to mention that you also need to import the `DatabaseCreatedEvent` as it's referenced in the sample code below.


marceloneppel (approved):
LGTM! Thanks for catching that, @mthaddon!

paulomach (approved):
lgtm

PR #27, opened 2022-11-17
add is_database_created function

delgod:


lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -418,6 +418,23 @@ def fetch_relation_data(self) -> dict:
...
+
+        Returns:
+            True or False
+        """
+        for relation in self.relations:
+            if (
+                "username" in relation.data[relation.app]
+                and "password" in relation.data[relation.app]
+            ):
+                return True
```
beliaev-maksim, line 435: does not it assume that there is only one relation then?
it will always exit after the first check

delgod:
The "database_requires" and "database_provides" libraries are now deprecated in favour of "data_interfaces", which should provide a more comprehensive framework to extend to other technologies beside databases. On this migration, the signature of "DatabaseRequires" and "DatabaseProvides" did not change. 
PR should be created for a new library.

PR #28, opened 2022-11-18
First version of extend data-platform-libs to support Kaka and Zookeeper

welpaolo:
This is a first draft that extend data-platform-libs to support Kafka and Zookeeper


lib/charms/data_platform_libs/v0/database_provides.py
```diff
@@ -121,10 +179,14 @@ class DatabaseEvents(CharmEvents):
 deleted - key that were deleted"""
 
 
-class DatabaseProvides(Object):
-    """Provides-side of the database relation."""
+class DataProvidesMeta(type(Object), type(ABC)):
```
deusebio, line 182: I would define this for both Provides and Requires. We might call this class just "_AbstractMetaclass"

```
from abc import ABCMeta
from ops.framework import _Metaclass

class _AbstractMetaclass(ABCMeta, _Metaclass):
    pass
```

lib/charms/data_platform_libs/v0/database_provides.py
```diff
@@ -277,40 +380,97 @@ def set_replset(self, relation_id: int, replset: str) -> None:
...
             relation_id: the identifier for a particular relation.
-            tls_ca: TLS certification authority.
+            bootstrap_server: the bootstrap server address.
         """
-        self._update_relation_data(relation_id, {"tls_ca": tls_ca})
+        self._update_relation_data(relation_id, {"bootstrap-server": bootstrap_server})
 
-    def set_uris(self, relation_id: int, uris: str) -> None:
-        """Set the database connection URIs in the application relation databag.
+    def set_endpoints(self, relation_id: int, endpoints: str) -> None:
```
deusebio, line 426: Do we also need this method? Beside the description says "bootstrap server". I believe for Kafka users bootstrap-servers is a bit more familiar and specific than "endpoints". I would go with bootstrap-server only

deusebio: Beside, I see it now that the method is already present in the parent class

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -496,7 +630,129 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
...
+        if self.extra_user_roles:
+            self._update_relation_data(
+                event.relation.id,
+                {
+                    "topic": self.topic,
+                    "extra-user-roles": self.extra_user_roles,
+                },
+            )
+        else:
+            self._update_relation_data(event.relation.id, {"topic": self.topic})
```
deusebio, line 659: I understand it was done this way in other parts of the code, but (I know it is a matter of taste) isn't it more elegant the tertiary formulation:

```
self._update_relation_data(
    {"topic": ..., "extra-user-role": ...} if self.extra_user_roles is not None else {"topic": ...}
)
```

Clearly not a big point, just a styling suggestion subjected to a matter of taste

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -496,7 +630,129 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
...
+            logger.info("topic created at %s", datetime.now())
+            self.on.topic_created.emit(event.relation, app=event.app, unit=event.unit)
+
+            # To avoid unnecessary application restarts do not trigger
+            # “endpoints_changed“ event if “topic_created“ is triggered.
+            return
+
+        # Emit an endpoints changed event if the Kakfa endpoints
+        # added or changed this info in the relation databag.
+        if "endpoints" in diff.added or "endpoints" in diff.changed:
```
deusebio, line 679: Uhm, I'm not sure this is a relevant event for kafka, given we are already tracking the bootstrap-server changed event

@marcoppenheimer  what do you think?

tests/integration/kafka-charm/metadata.yaml
```diff
@@ -0,0 +1,16 @@
...
+name: kafka
+description: |
+  Toy charm used to emulate Kafka in integration tests.
+summary: |
+  Charm used to mimic the Kakfa charm for test purpose only.
+
+
+peers:
+  kafka-peers:
+    interface: kafka-peers
```
deusebio, line 12: Is peers relation required? If not I would skip this, just to keep the charm minimal

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1271 @@
...
+— changed — keys that still exist but have new values.
+— deleted — keys that were deleted.
+"""
+
+
+class _AbstractMetaclass(ABCMeta, _Metaclass):
+    pass
+
+
+class BaseRequires(Object, ABC, metaclass=_AbstractMetaclass):
```
deusebio, line 905: DataRequires?

Just be consistent with the other class `DataProvides`

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1271 @@
...
+    return Diff(added, changed, deleted)
+
+
+class _AbstractMetaclass(ABCMeta, _Metaclass):
+    """Meta class."""
+
+    pass
+
+
+class DataProvides(Object, ABC, metaclass=_AbstractMetaclass):
```
deusebio, line 454: DataProvides or BaseProvides? I would go with DataProvides (not strong opinion on this) but be consistent throughtout

@marceloneppel what do you think?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1271 @@
...
+        only the leader unit can call it.
+
+        Args:
+            relation_id: the identifier for a particular relation.
+            connection_strings: database hosts and ports comma separated list.
+        """
+        self._update_relation_data(relation_id, {"endpoints": connection_strings})
+
+
+class BaseRequiresEvent(RelationEvent):
```
deusebio, line 723: Uhm, what about calling this event more like "AuthenticationEvent"?

The fields sounds a bit more like this

deusebio (commented):
The code looks very good and I like it a lot! Terrific job!

I only have one major comment, which should also prevent some minor mistakes (about having duplicate classes and inconsistent naming, see comments). 

I would rather organize the code in this way:

1. General functions (Diff, diff)
2. Abstraction (_AbstractMetaclass, DataProvides, DataRequires)
3. General Events (ExtraRoleEvent, AuthentioncationEvent)
4. Custom Implementation Sections
     4.1 Database (Database*Events, DatabaseProvider, DatabaseRequirer)
     4.2 Kafka (Kafka*Events, KafkaProvider, KafkaRequirer)
     4.3 Zookeeper (Zookeeper*Events, ZookeeperProvider, ZookeeperRequirer)    

I know it sounds a small point but I believe this would improve readability of the library, plus adding a new technology should be just a section, instead of being scattered around

tests/unit/test_data_interfaces.py
```diff
@@ -0,0 +1,955 @@
...
+            assert all(
+                isinstance(captured_event, event["event"]) for captured_event in captured_events
+            )
+
+            # Test that the remote unit name is available in the event.
+            for captured in captured_events:
+                assert captured.unit.name == "database/0"
+
+
+class TestKakfaRequires(DataRequirerBaseTests, unittest.TestCase):
```
marceloneppel, line 783: Should we have also a `TestZookeeperRequires` class?

marceloneppel (commented):
Awesome design and implementation Paolo! I don't have any comments about the way it was designed. I left some questions and some minor changes in the code and docs.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1271 @@
...
+        # added or changed this info in the relation databag.
+        if "endpoints" in diff.added or "endpoints" in diff.changed:
+            # Emit the default event (the one without an alias).
+            logger.info("endpoints changed on %s", datetime.now())
+            self.on.endpoints_changed.emit(event.relation, app=event.app, unit=event.unit)
+            return
+
+        # Emit a read only credential changed event if the Zookeeper credentials
+        # changed this info in the relation databag.
+        if "username" in diff.changed or "password" in diff.changed:
```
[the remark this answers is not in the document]
deusebio: Uhm, that's actually a good point. Beside, I might see the use-case to rotate the password, but not the use-case to rotate the username. Rotating by clearing the relation could be an option too. 

@marceloneppel do you have references for this design you talk about? like a spec doc or any trace of that discussion?

tests/unit/test_data_interfaces.py
deusebio: Actually, we went a bit ahead with the zookeeper implementation, as we should still open a pr on the https://github.com/canonical/charm-relation-interfaces repo to agree on the signature for the relation.

We don't need the Zookeeper interface yet and we could also post-pone the merge of this feature once we have agreed on the interface signature

Maybe it would actually be best to stash the bit for zookeeper in a separate branch and merge it only once we have decided on the interface. Together with implementing the tests

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
marceloneppel: I don't remember if some design document was written. I just remember about talking to Mykola about it. So maybe we should clarify this with him.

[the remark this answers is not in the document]
marceloneppel: [This](https://chat.canonical.com/canonical/pl/hummpwzeg3dc78mfqg8sesm7bw) is a conversation about this topic.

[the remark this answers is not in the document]
deusebio: Thanks a lot for this!!!! Awesome memory!! I believe we should then drop the CredentialChanged event and keep consistent UX across interfaces. 

We can open an issue though, if we see a use-case for this, to follow up the discussion and see whether we want to create such event across data platforms

tests/integration/kafka-charm/src/charm.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        # Default charm events.
+        self.framework.observe(self.on.start, self._on_start)
+
+        # Charm events defined in the Kafka Provides charm library.
+        self.kafka_provider = KafkaProvides(self, relation_name="kafka_client")
+        self.framework.observe(self.on[PEER].relation_created, self._on_peer_relation_changed)
```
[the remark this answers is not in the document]
marceloneppel: In this context peer relation is useful to share some data with the actions that the charm provides for the tests. I think it's ok to keep it as is now (to not use stored state - that [is not recommended in most of the cases](https://juju.is/docs/sdk/stored-state-uses-limitations)). Otherwise we wouldn't have the data available for the actions.

It's a good approach to have the peer data bags as the secret store until we move to Juju secrets.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1271 @@
...
+        Args:
+            relation_id: the identifier for a particular relation.
+            replset: replica set name.
+        """
+        self._update_relation_data(relation_id, {"replset": replset})
+
+    def set_uris(self, relation_id: int, uris: str) -> None:
+        """Set the database connection URIs in the application relation databag.
+
+        MongoDB, Redis, OpenSearch and Kafka only.
```
marceloneppel, line 620: Should Kafka be removed from this docstring as with this PR we'll have specific Kafka fields (bootstrap server and zookeeper uris)?

tests/unit/test_data_interfaces.py
welpaolo: Ok I will remove Zookeeper for the moment and put it in a separate branch

lib/charms/data_platform_libs/v0/data_interfaces.py
welpaolo: Ok it makes sense

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1081 @@
...
+application related data.
+
+#### Requirer Charm
+
+```python
+
+from charms.data_platform_libs.v0.data_interfaces import (
+    BootstrapServerChangedEvent,
+    KafkaRequires,
+    KafkaCredentialsChangedEvent,
```
deusebio, line 196: This does not exist anymore, does it?

tests/integration/application-charm/src/charm.py
```diff
@@ -10,19 +10,24 @@
 
 import logging
 
-from charms.data_platform_libs.v0.database_requires import (
+from charms.data_platform_libs.v0.data_interfaces import (
+    BootstrapServerChangedEvent,
     DatabaseCreatedEvent,
     DatabaseEndpointsChangedEvent,
     DatabaseRequires,
+    KafkaCredentialsChangedEvent,
```
deusebio, line 18: This does not exists, does it?

tests/integration/application-charm/src/charm.py
```diff
@@ -167,6 +188,29 @@ def _on_cluster2_endpoints_changed(self, event: DatabaseEndpointsChangedEvent) -
         """Event triggered when the read/write endpoints of the database change."""
         logger.info(f"cluster2 endpoints have been changed to: {event.endpoints}")
 
+    def _on_kafka_credentials_changed(self, event: KafkaCredentialsChangedEvent):
```
deusebio, line 191: this does not exists anymore, does it?

tests/integration/application-charm/src/charm.py
welpaolo: Good catch!

deusebio (approved):
Good to go!

PR #38, opened 2023-01-26
[DPE-1134] Structured Config Library Implementation

deusebio:


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
welpaolo, line 1: I think we need to increase the `LIBPATCH `

lib/charms/data_platform_libs/v0/database_provides.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
welpaolo, line 1: Same as before

lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
welpaolo, line 1: Same here

lib/charms/data_platform_libs/v0/s3.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
welpaolo, line 1: Here also

requirements.txt
```diff
@@ -1 +1,2 @@
 ops >= 2.0.0
+pydantic>=1.10
```
welpaolo, line 2: @deusebio  Do we need to declare the `pydantic` dependencies in the library ?

welpaolo (approved):
Really nice library, I was able to the test with charm configurations and it worked really well!

requirements.txt
deusebio: YES, in the end it is really a matter of saying 

```
PYDEPS = ["pydantic>=1.10"]
```

zmraul (approved):
This is an interesting implementation!
I'm not very familiar with Pydantic or some of the approaches on the code, so take the approval with a grain of salt :)

lib/charms/data_platform_libs/v0/database_provides.py
deusebio: I'd skip this because it is deprecated. We should not really touch these files anymore I suppose

lib/charms/data_platform_libs/v0/database_requires.py
deusebio: I'd skip this because it is deprecated. We should not really touch these files anymore I suppose

marcoppenheimer (approved):

PR #40, opened 2023-01-30
[MISC] is_resource_created method

deusebio:
PR to supersed #27

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -559,8 +559,8 @@ class DataRequires(Object, ABC):
         when outside an event callback.
 
         Args:
-            relation_id: int, optional When provided the checks is done only for the id provided,
-                otherwise the checks is done for all relations
+            relation_id (int, optional): When provided the check is done only for the relation id provided,
+                otherwise the check is done for all relations
 
         Returns:
             True or False
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -462,10 +462,10 @@ class DataRequires(Object, ABC):
     """Requires-side of the relation."""
 
     def __init__(
-        self,
-        charm,
-        relation_name: str,
-        extra_user_roles: str = None,
+            self,
+            charm,
+            relation_name: str,
+            extra_user_roles: str = None,
     ):
         """Manager of base client relations."""
         super().__init__(charm, relation_name)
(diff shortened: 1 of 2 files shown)


delgod (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -546,6 +546,23 @@ def relations(self) -> List[Relation]:
...
+
+        Returns:
+            True or False
+        """
+        for relation in self.relations:
+            if (
+                "username" in relation.data[relation.app]
+                and "password" in relation.data[relation.app]
+            ):
+                return True
```
beliaev-maksim, line 563: how will it work for clusters?

can we check per cluster?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -546,6 +546,23 @@ def relations(self) -> List[Relation]:
         """The list of Relation instances associated with this relation_name."""
         return list(self.charm.model.relations[self.relation_name])
 
+    def is_resource_created(self) -> bool:
```
welpaolo, line 549: I propose to change this function to take as input the `relation_id` since multiple relations can be established. 
```python
def is_resource_created(self, relation_id:int) -> bool:
    assert relation_id in relation.data[relation.app]
    return "username" in relation.data[relation.app][relation_id] and "password" in relation.data[relation.app][relation_id]
```

welpaolo (changes requested):

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: uhm, interesting. What about using this business logic: if relation_id is not specified then resource in ALL relations must exists to return true. If the relation_id is provided, we only check for the given relation. I believe this provides a cleaner ux across cases

```
def _is_resource_created_for_relation(relation: Relation):
    return "username" in relation.data[relation.app][relation_id] and \
           "password" in relation.data[relation.app][relation_id]

def is_resource_created(self, relation_id: Optional[int]) -> bool:
    if relation_id:
        return _is_resource_created_for_relation(self.relations[relation_id])
    else:
        return all([_is_resource_created_for_relation(relation) for relation in self.relations])
```

beliaev-maksim: > if relation_id is not specified then resource in ALL relations must exists to return true. If the relation_id is provided, we only check for the given relation. 

I like it

natalian98 (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -546,6 +546,32 @@ def relations(self) -> List[Relation]:
...
+
+    def is_resource_created(self, relation_id: Optional[int] = None) -> bool:
+        """Check if the resource has been created.
+
+        This function can be used to check if the Provider answered with data in the charm code
+        when outside an event callback.
+
+        Args:
+            relation_id: int, optional When provided the checks is done only for the id provided,
+                otherwise the checks is done for all relations
```
beliaev-maksim, line 563:
```suggestion
        Args:
            relation_id (int, optional): When provided the check is done only for the relation id provided,
                otherwise the check is done for all relations
```

beliaev-maksim (changes requested):

beliaev-maksim (approved):

welpaolo (approved):

Issue #43, opened 2023-02-10
`is_resource_created` is returning `True` on relation-broken event


welpaolo:
### Bug Description

The function is `is_resource_created` returns True when it it called on `relation-broken` event.

### To Reproduce

N/A

### Environment

N/A

### Relevant log output

```shell
N/A
```

Issue #42, opened 2023-02-10
is_resource_created should return False where there is no relation


deusebio:
### Bug Description

When there are no relations, `is_resource_created` returns `True` since

```
all([]) == True
```

The correct UX that I would feel more intuitive is:

* when calling `is_resource_created()` with no relations, we return False
* when calling `is_resource_created(relation_id)` with no relation, we raise an exception 

### To Reproduce

N/A

### Environment

N/A

### Relevant log output

```shell
N/A
```

PR #44, opened 2023-02-11
[ISSUE-42;ISSUE-43] is_resource_created improvements and fixes

deusebio:
Addressing #42 and #43

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -501,12 +501,6 @@ class DataRequires(Object, ABC):
             a dict of the values stored in the relation data bag
                 for all relation instances (indexed by the relation ID).
         """
-        if "-relation-broken" in os.environ.get("JUJU_HOOK_NAME", ""):
-            # read more in https://bugs.launchpad.net/juju/+bug/1960934
-            raise RuntimeError(
-                "`fetch_relation_data` cannot be used in `*-relation-broken` events"
-            )
-
         data = {}
         for relation in self.relations:
             data[relation.id] = {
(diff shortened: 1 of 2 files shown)


welpaolo (changes requested):
This improvement looks really good. I will also fix the function `fetch_relation_data` and change the line 504 in this way:
``` python
if f"{self.relation_name}-relation-broken" in os.environ.get("JUJU_HOOK_NAME", ""):
```
This will avoid the raise of the runtime exception if the `relation-broken` event is not related to the database.

delgod (approved):

beliaev-maksim:
@deusebio @welpaolo @delgod 
so, you removed in this PR check for relation broken. But then it is not really clear what user did wrong. eg without reading docstring of the function you will never know

could this be improved?

> This will avoid the raise of the runtime exception if the relation-broken event is not related to the database.

was not check checking the name of the database relation to ensure it is only related to database relations?

deusebio:
I'm not sure I fully understand what you mean. IIRC, here the issue is that there may exist multiple relations under the same relation name, and `fetch_relation_data` fetches the databags for all of them, indexed by the relation id. 

Now, previously in the `relation_broken_event`, we were raising an exception in any relation broken event, which it is a bit too conservative, as people may still want to access the data for the relations other than the one tearing down. Indeed, this was a use case that some people stumbled upon (OSM team if I recall correctly). I remembered that there was also a somewhat inconsistent behaviour for accessing relation data in integration tests vs unittests (see MM thread [here](https://chat.charmhub.io/charmhub/pl/dcj35cm1s38e5jbz8qnma9b3ew)), but this might be a different story. 

So, here we basically just fetch the data that can be fetched. If the databag of the relation broken is not available, the dictionary would just NOT have the corresponding relation id key rather than raising an exception. I don't think the users would do anything wrong to use `fetch_relation_data` in `relation_broken_event`. Just there may not be some data there. Also, If they try to `self.fetch_relation_data()[relation_id]` for relations that had been removed, they will get an exception anyway.

beliaev-maksim:
@deusebio 
knowing the context it makes sense :)
however, during execution it might be confusing for users :)

Issue #45, opened 2023-02-16
`TypeError: metaclass conflict` encountered when using data_interfaces charm library


NucciTheBoss:
### Bug Description

My charm `slurmdbd-operator` fails to install because the error `TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases` is thrown when importing the __data_interfaces__ charm library.

Looking through the charm library, the issue is being caused by the `DataProvides` and `DataRequires` base classes trying to inherit from both `ops.framework.Object` and `abc.ABC`. `ops.framework.Object` and `abc.ABC` use different metaclasses so conflicts are encountered when trying to define `DataProvides` and `DataRequires` at charm runtime.

### To Reproduce

1. `git clone git@github.com:NucciTheBoss/slurmdbd-operator.git@mysql-migration`
2. `cd slurmdbd-operator && charmcraft -v pack`
3. `juju deploy ./slurmdbd_ubuntu-20.04-amd64_centos-7-amd64.charm slurmdbd`
4. See error once install hook is executed.

### Environment

I am running Juju locally using my LXD server hosted on my workstation. I am using `v0.6` of the __data_interfaces_ charm library. I am using Juju version 3.1 and my slurmdbd-operator is current using Ubuntu 20.04 LTS as its base.

### Relevant log output

```shell
unit-slurmdbd-1: 14:57:56 INFO juju.worker.uniter found queued "install" hook                                                                                                                                      
unit-slurmdbd-1: 14:57:56 WARNING unit.slurmdbd/1.install Traceback (most recent call last):                                                                                                                       
unit-slurmdbd-1: 14:57:56 WARNING unit.slurmdbd/1.install   File "./src/charm.py", line 12, in <module>                                                                                                            
unit-slurmdbd-1: 14:57:56 WARNING unit.slurmdbd/1.install     from charms.data_platform_libs.v0.data_interfaces import (                                                                                           
unit-slurmdbd-1: 14:57:56 WARNING unit.slurmdbd/1.install   File "/var/lib/juju/agents/unit-slurmdbd-1/charm/lib/charms/data_platform_libs/v0/data_interfaces.py", line 354, in <module>                           
unit-slurmdbd-1: 14:57:56 WARNING unit.slurmdbd/1.install     class DataProvides(Object, ABC):                                                                                                                     
[...]

PR #46, opened 2023-02-17
[ISSUE-45] Update documentation and data_interface library dependency

deusebio:
Addressing #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,13 @@ which they may be imported and used as normal charms. For example:
 Following are the libraries available in this repository:
 
 - `data_interfaces` - Library to manage the relation for the data-platform products.
+- `data_models` - Library to introduce pydantic models for handling configuration, action parameters and databags.
 - `database_provides` - [DEPRECATED] Library that offers custom events and methods for provider-side of the
   relation (e.g. mysql)
 - `database_requires` - [DEPRECATED] Library that offers custom events and methods for requirer-side of the
   relation (e.g. wordpress)
+
+The libraries are compatible with `ops>=2.0`. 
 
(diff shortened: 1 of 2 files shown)


welpaolo (approved):
Looks good to me!

marceloneppel (approved):
LGTM!

README.md
```diff
@@ -21,11 +21,14 @@ which they may be imported and used as normal charms. For example:
 Following are the libraries available in this repository:
 
 - `data_interfaces` - Library to manage the relation for the data-platform products.
+- `data_models` - Library to introduce pydantic models for handling configuration, action parameters and databags.
 - `database_provides` - [DEPRECATED] Library that offers custom events and methods for provider-side of the
   relation (e.g. mysql)
 - `database_requires` - [DEPRECATED] Library that offers custom events and methods for requirer-side of the
   relation (e.g. wordpress)
 
+The libraries are compatible with `ops>=2.0`.
```
NucciTheBoss, line 30: It might be better to have this as the following:

```suggestion
> __Note:__ `data_interfaces` is not compatible `ops<=1.5.4`. It is compatible with only `ops>=2.0.0`. 
```

This way it is very clear to the reader that `data_interfaces` will not work if being used with an older version of ops, and if you want to use the library, then you need to be on at least ops==2.0.0. With how it is now, it is not clear that `data_interfaces` is not backwards compatible with older versions of ops.

NucciTheBoss (changes requested):

NucciTheBoss (approved):
LGTM! Thank you for putting in the work  @deusebio :smile:

PR #49, opened 2023-03-02
Bump juju to 2.9.38, libjuju to 2.9.38.1 and remove flake8 pinning

taurus-forever:
Update components to the latest versions in Juju 2.9 LTS.
The old pinning should no longer be necessary.


deusebio (approved):
super cool! I see that we are not yet using poetry! Should we do this @delgod ? I'm would up vote for this

delgod (approved):

welpaolo (approved):

PR #51, opened 2023-03-15
[MISC] Bug fix writing in the local unit instead of local app

deusebio:
I have noticed that the Requirer side write the "data" field (used for computing differences) in the local_unit instead of the local_app


marceloneppel (commented):
I think the local unit is used due to https://github.com/canonical/data-platform-libs/pull/12.

delgod:
closed because of https://github.com/canonical/data-platform-libs/pull/51#pullrequestreview-1343830312

PR #54, opened 2023-03-29
[DPE-1180] Replace icon with the one provided by PM/WEB design team

taurus-forever:
It is a first repo to update the icon, check it in stable, and copy to all other Data Platform charms.


taurus-forever (commented):
I am worried about 
1) the size, the old icon 1Kb, the new one 19,5Kb.
2) I see header `<?xml version="1.0" encoding="UTF-8" standalone="no"?>` inside SVG icon. Is it expected here?

icon.svg
```diff
@@ -1 +1,288 @@
...
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
taurus-forever, line 1: Is XML header expected in SVG icon?

taurus-forever: Tnx! Any ideas why XML header was missing for old SVG file?

dragomirp: Likely removed to reduce size (minification).

delgod (approved):

icon.svg
delgod: yes, we need vector

deusebio (approved):
LGTM

taurus-forever:
Aborting as we need a [vector](https://github.com/canonical/data-platform-libs/pull/54#discussion_r1153996847) SVG icon.

PR #63, opened 2023-06-09
fix ModelError exception

delgod:
Using Juju 3.2.0 and data_interfaces 0.15, if the developer tries to access the relation data early in the charm's life (i.e. in the constructor), then in some cases, a ModelError is thrown and not yet caught.

This occurs most obviously when scaling an application - the peer relation joined hook fires very early in the new unit's life, and if _is_relation_active is called at this point, a ModelError can be thrown.

fixes https://pastebin.canonical.com/p/zKJPQ8B3gy/


marceloneppel (approved):
LGTM!

welpaolo (approved):
Good to go!

PR #64, opened 2023-06-21
chore: core structure of base DataUpgrade lib and handlers for pre upgrade-charm

marcoppenheimer:
##### `feat: add dependency conflict validators`
- `build_complete_sem_ver()` converts string versions into full major.minor.patch semantic versions, padding with 0s where not specified
    - etc `1.3` --> `[1, 3, 0]`
- All verification functions called in `verify_inequality_requirements` wrapper, returning boolean if passed arg version meets passed arg requirement
- Involved functions support version definition specification used by [`poetry`](https://python-poetry.org/docs/dependency-specification/), supporting:
    - [Caret](https://python-poetry.org/docs/dependency-specification/#caret-requirements)
    - [Tilde](https://python-poetry.org/docs/dependency-specification/#tilde-requirements)
    - [Wildcard](https://python-poetry.org/docs/dependency-specification/#wildcard-requirements)
    - [Inequality](https://python-poetry.org/docs/dependency-specification/#inequality-requirements)
        - NOTE - Choosing to support only `>` and `>=` for the time being, as an 'upgrade' implies increasing in version
##### `feat: strict typing of input dependencies with DependencyModel`
- Charm authors can define a dict of arbitrarily named keys in their charm with value of the dependencies for the current version of the charm, passing it in to the constructor of `DataUpgrade`
- The Pydantic model evaluates these specified dependencies for compliance to the above-mentioned specification, and whether specified `version` is compatible with specified `upgrade_supported`
- **REVIEW NOTE** - Idea is to load back the persisted dependencies dict from relation data into the same type as the defined model, and use something like:
    ```python
    if not old_deps >> new_deps:
        raise
    ```
    - Use of the bitshift operator is fairly weird, but I thought it was actually quite a nice override, very much welcome opinions here
##### `feat: custom Exceptions for various failure cases`
- `UpgradeError` base-class
- `ClusterNotReadyError` to be raised during `pre_upgrade_check()` by charm authors
- `UpgradeFailedError` to be raised during `upgrade_granted_event()` handlers by charm authors
##### `feat: DataUpgrade object for handling and emitting of upgrade events`
- Persists charm-author passed `dependencies` to relation data during `upgrade-relation-created` handler 
- Exposes properties for use by charm authors for `state`, `stored_dependencies`, `upgrade_stack`
- Abstractmethods for `pre_upgrade_check` and `build_upgrade_stack`, to be overridden by charm authors
- `pre_upgrade_check` handler does the following:
    - Fails if not leader
    - Fails if **ANY** unit is already in the process of upgrading (by checking `state`)
    - Runs charm-author defined `pre_upgrade_check`
    - Runs charm-author defined `build_upgrade_stack`, writing it to relation data
    - Sets all units to `state=waiting`
##### `style: add ruff`
- Basic ruff `pyproject.toml` specification, in preparation for a future PR to this repo to replace `flake8` for all libraries

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -308,7 +308,7 @@ class DependencyModel(BaseModel):
     def upgrade_supported_validator(
         cls, values
     ) -> dict[str, dict[str, Dependency] | str | Dependency]:
-        """Validates specified `version` is not behind specified `upgrade_supported` requirement."""
+        """Validates specified `version` meets `upgrade_supported` requirement."""
         if not verify_requirements(
             version=values.get("version"), requirement=values.get("upgrade_supported")
         ):
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -389,9 +389,7 @@ class DataUpgrade(Object, ABC):
         self.framework.observe(
             self.charm.on[relation_name].relation_changed, self._on_upgrade_changed
         )
-        self.framework.observe(
-            self.charm.on[relation_name].relation_changed, self._on_upgrade_changed
-        )
+        self.framework.observe(self.charm.on.upgrade_charm, self._on_upgrade_charm)
 
         # actions
         self.framework.observe(
@@ -431,6 +429,27 @@ class DataUpgrade(Object, ABC):
         return (
             json.loads(self.peer_relation.data[self.charm.app].get("upgrade-stack", "[]")) or None
         )
+
+    @property
+    def upgrading_units(self) -> tuple[str, str] | None:
+        """Check whether any peer units are currently upgrading.
+
+        Returns:
+            Tuple of upgrading unit name and state.
+        """
+        if not self.peer_relation:
+            return None
+
+        for unit in set([self.charm.unit] + list(self.peer_relation.units)):
+            if (current_state := self.peer_relation.data[unit].get("state")) in {
+                "ready",
+                "upgrading",
+                "completed",
+                "failed",
+            }:
+                return (unit.name, current_state)
+
+        return None
 
     @abstractmethod
     def pre_upgrade_check(self) -> None:
@@ -472,15 +491,9 @@ class DataUpgrade(Object, ABC):
             return
 
         # checking if upgrade in progress
-        for unit in set([self.charm.unit] + list(self.peer_relation.units)):
-            if (current_state := self.peer_relation.data[unit].get("state")) in {
-                "ready",
-                "upgrading",
-                "completed",
-                "failed",
-            }:
-                event.fail(f"{unit} is in {current_state} state and is currently upgrading.")
-                return
+        if unit := self.upgrading_units:
+            event.fail(f"{unit[0]} is in {unit[1]} state and is currently upgrading.")
+            return
 
         try:
             logger.info("Running pre-upgrade-check...")
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -3,11 +3,15 @@ from ops.charm import (
     ActionEvent,
     CharmBase,
     RelationChangedEvent,
+    RelationCreatedEvent,
     UpgradeCharmEvent,
 )
 from ops.framework import EventBase, Object
-from overrides import override
+from ops.model import Relation
 import logging
+import pydantic
+from pydantic import BaseModel, ValidationError
+import re
 
 
 # The unique Charmhub library identifier, never change it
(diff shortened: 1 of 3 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,517 @@
...
+        Example Usage:
+            ```python
+            old_deps: DependencyModel = foo  # from persisted relation data
+            new_deps: DependencyModel = bar  # from charm
+
+            if not old_deps >> new_deps:
+                raise
+            ```
+        """
+        raise NotImplementedError
```
marcoppenheimer, line 314: **question:** What do people think of this operation? It's fairly weird so I'm happy to remove it if it is undesirable.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,517 @@
...
+                "ready",
+                "upgrading",
+                "completed",
+                "failed",
+            }:
+                event.fail(f"{unit} is in {current_state} state and is currently upgrading.")
+                return
+
+        try:
+            self.pre_upgrade_check()
```
deusebio, line 489: **super-big niptick** maybe it is useful to log something out. Since upgrade can take some time, I'm thinking of the use-case where the user issues the action and then they don't know what's going on. Of course this should be up to charm authors when they implement the pre_upgrade_check provides some meaningful log during the process. But maybe one log here may not hurt as well

deusebio (approved):
Overall, it looks good. I only have a few minor points and considerations

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,517 @@
...
+            upgrade_stack = self.build_upgrade_stack()
```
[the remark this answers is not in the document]
marcoppenheimer: It can fail, which is why it's caught by the `except Exception as e:`.

I would ideally like charm authors to raise some exception inherited from `UpgradeError` here, but as it's their own implementation, I've kept the catching of it generic.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,518 @@
...
+        # checking if upgrade in progress
+        for unit in set([self.charm.unit] + list(self.peer_relation.units)):
+            if (current_state := self.peer_relation.data[unit].get("state")) in {
+                "ready",
+                "upgrading",
+                "completed",
+                "failed",
+            }:
+                event.fail(f"{unit} is in {current_state} state and is currently upgrading.")
+                return
```
marceloneppel, line 483: Maybe expose a similar check as a property (for the current unit only unit only), like `state`, to avoid repeating the check for all the states in charm code? (I mean the idea from the spec about something like `upgrade_in_progress` property to use when deferring other events until the upgrade is complete).

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,518 @@
...
+        for unit in set([self.charm.unit] + list(self.peer_relation.units)):
+            self.peer_relation.data[unit].update({"state": "waiting"})
+
+    # TODO - implement
+    def _on_upgrade_changed(self, event: RelationChangedEvent):
+        """Handler for `upgrade-relation-changed` events."""
+        raise NotImplementedError
+
+    # TODO - implement
+    def _on_upgrade_charm(self, event: UpgradeCharmEvent):
```
marceloneppel, line 516: Should it be called in the above duplicate observe call?

marceloneppel (commented):
Great work, Marc! One thing that I couldn't find in the code is the `upgrade_granted_event` being observer and/or fired.

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Yup! My bad there, will add.

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Done!

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Done!

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+
+    def __init__(self, message: str, cause: str, resolution: str):
+        super().__init__(message, cause=cause, resolution=resolution)
+
+
+# --- CUSTOM EVENTS ---
+
+
+class UpgradeGrantedEvent(EventBase):
```
marceloneppel, line 364: Did you see my last comment about `UpgradeGrantedEvent `? It's not being fired or observed.

marcoppenheimer: It won't be fired yet, as that firing will be handled in the handling of `upgrade-charm` event, which is not part of this PR. 

The 'handling' of that event is for the charm authors to implement in their charms, maybe in `charm.py`, where they can restart snap services, sync data etc.

marceloneppel: Thanks for details, Marc.

marceloneppel (approved):

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,531 @@
...
+        raise NotImplementedError
+
+
+# --- CUSTOM EXCEPTIONS ---
+
+
+class UpgradeError(Exception):
+    """Base class for upgrade related exceptions in the module."""
+
+    def __init__(self, message: str, cause: str | None, resolution: str | None):
```
paulomach, line 320: Union with `|` requires python 3.10 - I believe we need to be 3.8 compatible still (vm on focal), so use `typing.Union` instead.
If so, it would also apply to lines 344,400, 405 and 413

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,531 @@
...
+        """Handler for `upgrade-relation-created` events."""
+        if not self.charm.unit.is_leader() or not self.peer_relation:
+            return
+
+        # persisting dependencies to the relation data
+        self.peer_relation.data[self.charm.app].update(
+            {"dependencies": json.dumps(self.dependency_model.dict())}
+        )
+
+    def _on_pre_upgrade_check_action(self, event: ActionEvent):
```
paulomach, line 483: on the k8s in-place upgrades we propose to use the lib partially, that is, the dependency validation and pre-upgrade-check-action (as a concept).
Having a call to `self.build_upgrade_stack()` is not necessary on k8s but prompted me if is the case to pop (as in a separated module/file) the dependency validation code from the rest.

paulomach: btw, not critical since it's simple to workaround

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Who is supporting 3.8? I was (perhaps wrongly?) under the impression we were all on jammy.

marceloneppel: PgBouncer and probably mysql router charms on VM are on both focal and jammy because they when related as subordinate charms they need to be in the same base as the application charm (in some cases is focal).

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -0,0 +1,531 @@
...
+        class KafkaDependenciesModel(BaseModel):
+            kafka_charm: DependencyModel
+            kafka_service: DependencyModel
+
+        deps = {
+            "kafka_charm": {
+                "dependencies": {"zookeeper": ">5"},
+                "name": "kafka",
+                "upgrade_supported": ">5",
+                "version": "10",
```
paulomach, line 253: I've just remind me of one thing, suggestions on a way to retrieve the charm version/revision?

marcoppenheimer: Hard-code it. It can't be done programmatically.

paulomach (approved):
gorgeous

PR #67, opened 2023-06-22
DPE-1180 Fix display name: s/display_name/display-name/ + add docs

taurus-forever:
The incorrect option name used here historically, see commit 36a7a848b892ec46a982f290fee9ca519876d75e.
Specification: https://discourse.charmhub.io/t/file-metadata-yaml/5213#heading--display-name

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 
 name: data-platform-libs
-display_name: Data Platform Libs
+display-name: Data Platform Libs
 summary: Data platform libraries.
 description: |
   This library is a uniform interface to a selection of common database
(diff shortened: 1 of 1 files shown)


metadata.yaml
```diff
@@ -12,6 +12,7 @@ description: |
   The charm provides libraries for both bare-metal/virtual-machines and K8s.
 
   NOTE: This charm is not meant to be deployed
+docs: https://discourse.charmhub.io/t/data-platform-libs-documentation/10955
```
deusebio, line 15: I have seen the Discourse topic, but that one does not quite have any section (apart from libraries, which just link to charmhub). Maybe, until we have some content, we can just stay with parsing the "description" field in the metadata for the charmhub apperance. 

Anyhow, it is not too important or critical. If I had to choose, I would rather drop this line until we have a bit more meaningful content, but I don't feel very strongly about this

deusebio (approved):
Just a suggestion, but LGTM either way

metadata.yaml
taurus-forever: The goal is to mark charm "Listed". To list it it have to match "Definition of Great" (see the slide 15 there).
So, we have to provide a link to Discourse, IMHO. Good time to add basic parts, meanwhile I have marked those sections as "Coming soon".

delgod (approved):

PR #69, opened 2023-07-01
style: add pyright + ruff checks, tighten up lib typing

marcoppenheimer:
##### `style: add pyright type checks, ruff linting`
- Both runs on `tox -e lint`
- Validates only `src` and `lib`
- Ruff is faster, less configured, removes `flake8`
- Pyright's closed-source version (`Pylance`) is the default type checker that ships with the `ms-python.python` VSCode extension, and is very good


taurus-forever (approved):
LGTM, charmcraft is not happy a bit:
```
2023-07-01 17:32:12.936 :: 2023-07-01 17:32:12.087 ::    ::             error: can't find Rust compiler
....
2023-07-01 17:32:12.936 :: 2023-07-01 17:32:12.087 ::    ::         ERROR: Failed building wheel for maturin
```

Q: are we moving to ruff for entire team?

marcoppenheimer:
> LGTM, charmcraft is not happy a bit:
> 
> ```
> 2023-07-01 17:32:12.936 :: 2023-07-01 17:32:12.087 ::    ::             error: can't find Rust compiler
> ....
> 2023-07-01 17:32:12.936 :: 2023-07-01 17:32:12.087 ::    ::         ERROR: Failed building wheel for maturin
> ```
> 
> Q: are we moving to ruff for entire team?

I think we should! It's much faster than flake, fewer imports.

deusebio:
> I think we should! It's much faster than flake, fewer imports.

I agree (unless ruff conflicts with other tools with no clean solution for it) but given we don't have any standard set up yet, I would leave it up to the products/team/projects to decide whether to use it or not for the time being.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -265,7 +265,7 @@ class KafkaDependenciesModel(BaseModel):
         model = KafkaDependenciesModel(**deps)
 
         # exporting back validated deps
-        print(model.dict())
+        print(model.model_dump())
```
deusebio, line 268: Uhm, may be out of the scope for this PR, but we should not really use print. It would be best to use rather use logging :D

deusebio (approved):
LGTM

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Given that it's just a docstring, I think it's OK!

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -974,7 +1018,9 @@ def is_postgresql_plugin_enabled(self, plugin: str, relation_index: int = 0) ->
         try:
             with psycopg.connect(connection_string) as connection:
                 with connection.cursor() as cursor:
-                    cursor.execute(f"SELECT TRUE FROM pg_extension WHERE extname='{plugin}';")
+                    cursor.execute(
+                        "SELECT TRUE FROM pg_extension WHERE extname='{%s}';", (plugin,)
+                    )
```
marcoppenheimer, line 1023: ?

PR #70, opened 2023-07-06
DPE-2200 - chore: add post-upgrade handlers to upgrade lib

marcoppenheimer:
## TODO
- Add extensive module docstring
- Add unit tests for the event handlers

#### `feat: emit UpgradeGrantedEvent if unit is flagged as ready`
- Removed `state=waiting` from the design, as I forgot that Leader units can't set other Units' peer data. Instead all units set `state=ready` on `upgrade-charm` event
#### `feat: set available states to DataUpgrade class attr`
- Ordered by severity and expected execution order
- `cluster_state` grabs the lowest ordinal state from all unit states, i.e:
    - if ANY unit is `failed`, `cluster_state` will be `failed`
    - if ANY unit is `idle`, `cluster_state` will be `idle 
    - and so on...
- This helps with getting an overall picture of how far the cluster has progressed during an upgrade, or if it has failed
#### `feat: add substrate attr to DataUpgrade`
- Takes arguments `k8s` or `vm`
- If `k8s`, default `upgrade_stack` to all application unit.ids, ordered in increasing order. This is needed as proposed upgrade process for K8s charms will not include custom ordering of units in the StatefulSet
- As such, `build_upgrade_state` now not an `@abstractmethod`, raises `NotImplementedError` if `vm`
#### `feat: lazy-load upgrade-stack attr`
- When first called, will load from relation data and persist to object attr
- When popped from during handling of `upgrade-relation-changed` to get the next unit+state to upgrade, after re-setting it back to `DataUpgrade`, the lib persists the popped list to relation data
- In case the leader is the next unit after popping, the lib recurses `on_upgrade_changed` after doing this to ensure it gets the next event (as `relation-changed` events don't re-emit to the unit that changed it)
#### `feat: convenience methods for setting failed and completed unit upgrade state`
- Charm-authors should not be able to modify unit upgrade states that are used for coordinating events (`idle`, `ready`, `upgrading`)
- They can use the above methods during handling of `pre_upgrade_check` and `UpgradeGranted` event handlers to inform other units of their final states, triggering `upgrade-relation-changed` events on all other units to wake them up
    - If any unit state is set to `failed`, all units will get `upgrade-relation-changed`, and set their own state to `failed`
    - If a unit sets it's state to `completed`, all units will get `upgrade-relation-changed`, and the leader will pop the completed unit from the stack 
#### `feat: add can_upgrade method to DependencyModel`
- Compares two dependency models for base upgradability by comparing old `version` to new `upgrade_supported`
- Called implicitly on the application leader during `upgrade-charm` event handling through the `_upgrade_supported_check` method, which loops over all specified dependencies (e.g `snap`, `service`, `deb`...) and validates
#### `feat: Add VersionError and DependencyError exceptions`
- `VersionError` raised if `can_upgrade` fails for 'internal' versions (Snap --> Snap, Service --> Service etc)
- `DependencyError` can be raised by charm-authors during handling of `upgrade-granted` event if related applications do not meet requirements
#### `docs: various docstring tweaks to fit reStructuredText patterns`
- Added detailed usage instructions for `UpgradeGrantedEvent` for charm-authors to include in their handler implementations

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -39,7 +39,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 logger = logging.getLogger(__name__)
 


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -429,66 +479,127 @@ def stored_dependencies(self) -> Optional[BaseModel]:
...
+        if self.substrate == "k8s":
+            pass
+
+        raise NotImplementedError
+
+    @abstractmethod
+    def log_rollback_instructions(self) -> None:
+        """Sets charm state and logs out rollback instructions.
+
+        Called by all units when `state=failed` found during :meth:`_on_upgrade_changed`.
```
marceloneppel, line 570: Should we raise an `NotImplementedError` to force this method implementation? I know it's not something needed by the upgrade logic of the library, but it's so important for the user to know how to proceed.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -519,21 +638,108 @@ def _on_pre_upgrade_check_action(self, event: ActionEvent):
...
+            old_dep: DependencyModel = getattr(self.stored_dependencies, key)
+            new_dep: DependencyModel = getattr(self.dependency_model, key)
 
-    # TODO - implement
-    def _on_upgrade_charm(self, event: UpgradeCharmEvent):
+            if not old_dep.can_upgrade(dependency=new_dep):
+                raise VersionError(
+                    message="Versions incompatible - {key} {old_dep.version} can not be upgraded to {new_dep.version}",
+                    cause=f"Upgrades only supported from {key} versions satisfying requirement {new_dep.version}",
+                )
```
marceloneppel, line 660: I think it would be good to list all incompatibilities, to avoid next failures when the user try to upgrade again and only one of the incompatibilities was fixed (by, for example, an upgrade to one of the intermediate revisions that updates that first dependency that was reported as incompatible).

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -379,25 +414,32 @@ class UpgradeGrantedEvent(EventBase):
...
         self.relation_name = relation_name
+        self.substrate = substrate
 
         # events
         self.framework.observe(
             self.charm.on[relation_name].relation_created, self._on_upgrade_created
         )
         self.framework.observe(
-            self.charm.on[relation_name].relation_changed, self._on_upgrade_changed
+            self.charm.on[relation_name].relation_changed, self.on_upgrade_changed
```
marceloneppel, line 439: Not related to this PR, but I saw that `getattr(self.on, "pre_upgrade_check_action"), self._on_pre_upgrade_check_action` was failing, probably due to the need of it to be `getattr(self.charm.on, "pre_upgrade_check_action"), self._on_pre_upgrade_check_action`.

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: It's decorated with `@abstractmethod` which raises `TypeError` if you try to instantiate it without overriding the abstractmethod. 

e.g: 

```python
from abc import ABC, abstractmethod

class Base(ABC):
    def __init__(self, stuff: str):
        self.stuff = stuff

    @abstractmethod
    def get_stuff(self) -> str:
        pass
```

```python
class InheritedBase(Base):
    pass

ib = InheritedBase(stuff="some-stuff")
```
> TypeError: Can't instantiate abstract class InheritedBase with abstract method get_stuff

```python
class InheritedBase(Base):
    @override
    def get_stuff(self) -> None:
        print(self.stuff)

ib = InheritedBase(stuff="some-stuff")
ib.get_stuff()
```
> some-stuff

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: Great idea! I'll make that change now.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -429,66 +479,127 @@ def stored_dependencies(self) -> Optional[BaseModel]:
...
 
         Returns:
-            List of integeter unit.ids, ordered by upgrade order
-                e.g [5, 2, 4, 1, 3]
+            Iterable of integeter unit.ids, LIFO ordered in upgrade order
+                i.e `[5, 2, 4, 1, 3]`, unit `3` upgrades first, `5` upgrades last
+        """
+        # don't raise if k8s substrate, uses default statefulset order
+        if self.substrate == "k8s":
+            pass
```
paulomach, line 562: should you return here instead of pass

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -499,17 +610,25 @@ def _on_pre_upgrade_check_action(self, event: ActionEvent):
...
         try:
             logger.info("Running pre-upgrade-check...")
             self.pre_upgrade_check()
 
-            logger.info("Building upgrade stack...")
-            upgrade_stack = self.build_upgrade_stack()
-            logger.info(f"Built upgrade stack of {upgrade_stack}")
+            if self.substrate == "k8s":
+                logger.info("Building upgrade stack for K8s...")
+                built_upgrade_stack = sorted(
```
paulomach, line 623: tks! having this will be useful to control who's next, since the order is a given

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -519,21 +638,108 @@ def _on_pre_upgrade_check_action(self, event: ActionEvent):
...
+                self.set_unit_failed()
+                return
+
+        # all units sets state to ready
+        self.peer_relation.data[self.charm.unit].update({"state": "ready"})
+
+    def on_upgrade_changed(self, event: RelationChangedEvent) -> None:
+        """Handler for `upgrade-relation-changed` events."""
+        if not self.peer_relation:
+            event.defer()
```
paulomach, line 695: need to defer? a change to the peer relation will trigger the event anyway, won't it?

paulomach (approved):
left some nits. lgtm otherwise
core structure is nice
nothing obvious impairing k8s, will use it and 2-check

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -409,6 +451,14 @@ def peer_relation(self) -> Optional[Relation]:
         """The upgrade peer relation."""
         return self.charm.model.get_relation(self.relation_name)
 
+    @property
+    def peer_units(self) -> Iterable[Unit]:
```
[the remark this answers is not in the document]
marcoppenheimer: Done!

marceloneppel (approved):

PR #73, opened 2023-07-12
fix: various bugs

marcoppenheimer:
#### `fix: avoid recursion limit errors on leader`
- `[]` and `None` are both falsey in Python. When the final unit had been popped from the list, we would attempt to set that empty list to relation data by referencing the property. The property checks for `if not X: fetch`, getting the un-popped value from the relation data
- When we recursed on the leader, we'd get recursion limit errors, as it would never exit due to it just re-setting the original stack
- Fixed by specifically checking for `if self._upgrade_stack is None`
#### `fix: actually set default idle state on relation created`
- When unset, we'd handle relation events before upgrades had progressed
- Fixed, now we can check for `idle` and exit

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -593,16 +593,14 @@ class DataUpgrade(Object, ABC):
             event.defer()
             return
 
-        logger.info("Setting upgrade state to idle...")
+        # setting initial idle state needed to avoid execution on upgrade-changed events
         self.peer_relation.data[self.charm.unit].update({"state": "idle"})
 
-        if not self.charm.unit.is_leader():
-            return
-
-        logger.info("Setting charm dependencies to relation data...")
-        self.peer_relation.data[self.charm.app].update(
-            {"dependencies": json.dumps(self.dependency_model.dict())}
-        )
+        if self.charm.unit.is_leader():
+            logger.debug("Persisting dependencies to upgrade relation data...")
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -707,7 +708,7 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
         # all units sets state to ready
         self.peer_relation.data[self.charm.unit].update({"state": "ready"})
 
-    def on_upgrade_changed(self, event: RelationChangedEvent) -> None:
+    def on_upgrade_changed(self, event: EventBase) -> None:
```
paulomach, line 711: why change the event type?

paulomach (approved):
lgtm

lib/charms/data_platform_libs/v0/upgrade.py
marcoppenheimer: We have to call it from `upgrade-granted` events. If the Juju leader is the last unit to upgrade, it sets `state=completed`, but won't get the `upgrade-changed` event to remove itself from the stack and set `state=idle`. 

Because we're passing `UpgradeGranted` event - [example here](https://github.com/canonical/zookeeper-operator/blob/0f0be763f1fdc45aa5b499df07d08486b4f3fdb0/src/upgrade.py#L103-L104), we need to generalise the type.

marceloneppel (approved):

PR #74, opened 2023-07-12
Use `relation_created` instead of `relation_joined`

carlcsaposs-canonical:
The handlers should run when the relation is created, not when a remote unit joins the relation


paulomach (approved):
sensible

welpaolo (approved):
Looks good to me

taurus-forever (approved):

PR #75, opened 2023-07-13
chore: allow pre-upgrade-check action recovery on failed units

marcoppenheimer:


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -608,13 +608,23 @@ def _on_pre_upgrade_check_action(self, event: ActionEvent) -> None:
             event.fail(message="Could not find upgrade relation.")
             return
 
-        if not self.charm.unit.is_leader():
-            event.fail(message="Action must be ran on the Juju leader.")
+        # if check ran after failed upgrade, set recovery state to unblock failed unit
+        if self.peer_relation.data[self.charm.unit].get("state") == "failed":
+            self.peer_relation.data[self.charm.unit].update({"state": "recovering"})
+            return
```
marceloneppel, line 614: Can the `failed` state be present in any other unit at this point? If so, I think the check from L617 will be evaluated to true in that case.

marcoppenheimer: Ah this PR is about to be replaced, I will close it!!

PR #77, opened 2023-07-17
[DPE-2269] Added k8s upgrade handler

marceloneppel:
## Issue
`UpgradeGrantedEvent ` is emitted when the charm is allowed to upgrade the workload, but on K8s it's already upgraded when the charm receives `UpgradCharmEvent` and the other events that are fired later. For K8s, we need a different event, that tells that the upgrade was finished in the current unit and Juju can proceed to the next unit.

## Solution
Created `UpgradeFinishedEvent ` to be used to tell Juju to upgrade the next unit (by, for example, changing the `partition` field from the StatefulSet rolling update strategy). Logic for that was kept in the charms themselves for now to let them do some additional logic if needed.

Conditionals were added to trigger `UpgradeRelationChangedEvent` in the current unit on K8s, which avoid losing the event when we have only part of the units in the new charm revision.
 
Added `PyYAML` to `charm-binary-python-packages` due to https://github.com/yaml/pyyaml/issues/601.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -6,6 +6,7 @@ parts:
   charm:
     charm-binary-python-packages:
       - pydantic
+      - PyYAML # https://github.com/yaml/pyyaml/issues/601
     build-packages:
       - rustc
       - cargo
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -6,7 +6,6 @@ parts:
   charm:
     charm-binary-python-packages:
       - pydantic
-      - PyYAML # https://github.com/yaml/pyyaml/issues/601
     build-packages:
       - rustc
       - cargo
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -703,6 +703,9 @@ class DataUpgrade(Object, ABC):
             event.fail(message="Nothing to resume, upgrade stack unset.")
             return
 
+        # Check whether this is being run after juju refresh was called
+        # (the size of the upgrade stack should match the number of total
+        # unit minus one).
         if len(self.upgrade_stack) != len(self.peer_relation.units):
             event.fail(message="Upgrade can be resumed only once after juju refresh is called.")
             return
@@ -866,6 +869,8 @@ class DataUpgrade(Object, ABC):
             return
 
         try:
+            # Use the unit number instead of the upgrade stack to avoid race conditions
+            # (i.e. the leader updates the upgrade stack after this hook runs).
             next_partition = unit_number - 1
             logger.debug(f"Set rolling update partition to unit {next_partition}")
             self._set_rolling_update_partition(partition=next_partition)
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -38,7 +38,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 PYDEPS = ["pydantic>=1.10,<2"]
 
@@ -400,6 +400,17 @@ class UpgradeGrantedEvent(EventBase):
     """
 
 
+class UpgradeFinishedEvent(EventBase):
+    """Used to tell units that they finished the upgrade.
+
+    Handlers of this event must meet the following:
+        - MUST trigger the upgrade in the next unit by, for example, decrementing the partition
+            value from the rolling update strategy
+        - MUST update unit `state` if the previous operation fails, calling
+            :class:`DataUpgrade.set_unit_failed`
+    """
+
+
 class UpgradeEvents(CharmEvents):
     """Upgrade events.
 
@@ -407,6 +418,7 @@ class UpgradeEvents(CharmEvents):
     """
 
     upgrade_granted = EventSource(UpgradeGrantedEvent)
+    upgrade_finished = EventSource(UpgradeFinishedEvent)
 
 
 # --- EVENT HANDLER ---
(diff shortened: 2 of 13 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -556,7 +569,7 @@ def build_upgrade_stack(self) -> List[int]:
         Called by leader unit during :meth:`_on_pre_upgrade_check_action`.
 
         Returns:
-            Iterable of integeter unit.ids, LIFO ordered in upgrade order
+            Iterable of integer unit.ids, LIFO ordered in upgrade order
```
marcoppenheimer, line 572: **praise:** Thank you!

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -597,6 +610,9 @@ def set_unit_completed(self) -> None:
 
         self.peer_relation.data[self.charm.unit].update({"state": "completed"})
 
+        if self.substrate == "k8s":
+            self.on_upgrade_changed(EventBase(self.handle))
+
```
marcoppenheimer, line 615: **todo:** I think we should remove this block.

It's a good idea to have the function do it (so charm authors / users don't have to remember to do it), but I think we should just be explicit with the documentation that - when handling `upgrade-granted` or `upgrade-finished` - if the unit completes, call `on_upgrade_changed`.

I'm not sure we're using it now, but I wonder if in the future we'd benefit from having an actual `event` to pass, rather than `EventBase(self.handle)`. Not sure.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -763,13 +781,31 @@ def on_upgrade_changed(self, event: EventBase) -> None:
...
+        raise NotImplementedError
+
+    def _on_upgrade_finished(self, event: UpgradeFinishedEvent) -> None:
+        """Handler for `upgrade-finished` events."""
+        # don't raise if vm substrate, only return
+        if self.substrate == "vm":
+            return
+
+        raise NotImplementedError
```
marcoppenheimer, line 811: **question:** Non-blocking, curious, what should charms do when handling this event? You mentioned in the PR description changin gthe `partition` field, do you have an example of that that you could share?

marcoppenheimer (changes requested):

lib/charms/data_platform_libs/v0/upgrade.py
paulomach: One would be [this](https://github.com/canonical/mysql-k8s-operator/blob/0f01b2f4d49a0196c4679014d1de123127c60c1e/src/upgrade.py#L200) , which just control triggering update for a follow up unit

marceloneppel: I think so. Let me move it to that hook, so we can probably remove the upgrade-finished hook.

lib/charms/data_platform_libs/v0/upgrade.py
marceloneppel: I think the library should help the charm authors in ways like this, to not forget to call needed parts of the workflow.

As per the other discussion in this PR, probably we can remove this block as we wouldn't have to trigger upgrade finished. I'm gonna test that.

charmcraft.yaml
```diff
@@ -6,6 +6,7 @@ parts:
   charm:
     charm-binary-python-packages:
       - pydantic
+      - PyYAML # https://github.com/yaml/pyyaml/issues/601
```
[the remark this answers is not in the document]
marceloneppel: Thanks for the update. Removed the workaround on https://github.com/canonical/data-platform-libs/commit/afc861bda4b622392e8cf85144b83c4765f85837.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -763,13 +781,53 @@ def on_upgrade_changed(self, event: EventBase) -> None:
...
+        if self.charm.unit == top_unit:
+            if self.substrate == "vm" and top_state in ["ready", "upgrading"]:
+                logger.debug(
+                    f"{top_unit} is next to upgrade, emitting `upgrade_granted` event and upgrading..."
+                )
+                self.peer_relation.data[self.charm.unit].update({"state": "upgrading"})
+                getattr(self.on, "upgrade_granted").emit()
+            if self.substrate == "k8s" and top_state == "completed":
+                logger.debug(
+                    f"{top_unit} has completed the upgrade, emitting `upgrade_finished` event..."
```
paulomach, line 793:
```suggestion
                    f"{top_unit.name} has completed the upgrade, emitting `upgrade_finished` event..."
```

marceloneppel: Good catch! Fixed on https://github.com/canonical/data-platform-libs/pull/77/commits/bf8b2bf0b047d70800a256f214e6aef36e1a827f.

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -765,11 +787,45 @@ def on_upgrade_changed(self, event: EventBase) -> None:
...
+                    logger.exception("Kubernetes StatefulSet patch failed")
+
+        That shouldn't happen for the last unit (the first that is upgraded). For that unit it
+        should be done through an action after the upgrade success on that unit is double-checked.
+        """
+        # don't raise if vm substrate, only return
+        if self.substrate == "vm":
+            return
+
+        raise NotImplementedError
```
marcoppenheimer, line 831: **thought:** I'm not sure it should block this PR, but leaving idea here.

I wonder if we should implement a k8s-only `abstractmethod` for something like `decrement_partition()`.

That way, we could still call it by default here - hopefully reducing the LoC charm authors need to implement - but we still have the flexibility of doing any other operations around that in the future. It also lines up with the decision you mentioned in an earlier comment about not adding dependencies, as the implementation can be up to the authors.

marcoppenheimer (approved):
Really sensible changes! Left some notes and ideas for future PRs, but nothing blocking.

lib/charms/data_platform_libs/v0/upgrade.py
marceloneppel: That's a great idea. I updated the PR in the last commits (https://github.com/canonical/data-platform-libs/pull/77/files/b231d5b827d32c86c8b9d4c2dae801b58ef68894..770245cdff09ac8053c90983a35ac4cce2c0cb70) to address that. I also added the `resume-upgrade` action (only for k8s) because it was mentioned in the `_on_upgrade_finished` method in the and would be another common code for the k8s charms.

I also added the unit test for those new methods (and fixed the existing ones to match the implementation from this PR).

paulomach (approved):
great

PR #81, opened 2023-07-25
DPE-2288 - feat: support rollbacks, add statuses

marcoppenheimer:
#### `feat: support managed rollback after failed upgrade`
- Units now stay in the same state they were before failing, not also failing
- Running `pre-upgrade-check` action after a cluster has failed sets `state=recovering` to the leader
- `upgrade-charm` modifies upgrade-stack to insert any units that were already popped + completed (i.e not the failed ones)
    - Failed units are inserted to the top of the existing stack 
- `upgrade-granted` now only emitted when all units are ready

#### `feat: add unit statuses during upgrade`
- `BlockedStatus` with optional `cause` passed by `set_unit_failed()` if unit fails during
- `BlockedStatus` with generic message if any other unit failed during
- `MaintenanceStatus` throughout
- `ActiveStatus` when setting back to `state=idle`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tests/integration/database-charm/lib/charms/data_platform_libs/v0/data_interface
 tests/integration/kafka-charm/lib/charms/data_platform_libs/v0/data_interfaces.py
 tests/integration/s3-charm/lib/charms/data_platform_libs/v0/s3.py'
 .vscode/
+.idea/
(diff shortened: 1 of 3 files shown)


paulomach (commented):
Added suggestions as the PR #82

paulomach (approved):
superb

marceloneppel (approved):

PR #82, opened 2023-07-27
adapt for k8s statuses

paulomach:
### pulled k8s support from main

bulk of changes are related to pulling already merged k8s support (PR #77)

### statuses changes for k8s

* `on_upgrade_charm` calls for `self._upgrade_supported_check()` for differing conditions for k8s and vm, setting `state` and statuses accordingly

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -544,6 +544,11 @@ class DataUpgrade(Object, ABC):
 
     @property
     def unit_states(self) -> list:
+        """Current upgrade state for all units.
+
+        Returns:
+            Unsorted list of upgrade states for all units.
+        """
         if not self.peer_relation:
             return []
 
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -435,7 +435,7 @@ class UpgradeEvents(CharmEvents):
 class DataUpgrade(Object, ABC):
     """Manages `upgrade` relation operators for in-place upgrades."""
 
-    STATES = ["recovery", "failed", "ready", "upgrading", "idle", "completed"]
+    STATES = ["recovery", "failed", "idle", "ready", "upgrading", "completed"]
 
     on = UpgradeEvents()  # pyright: ignore [reportGeneralTypeIssues]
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_upgrade.py
```diff
@@ -35,6 +35,8 @@
 GANDALF_ACTIONS = """
```
marceloneppel, line 35: Missing tests here? On main I see other tests that are not here anymore, like `test_resume_upgrade_action_succeeds_only_when_ran_at_the_right_moment`. Or is the idea to fix the conflicts on tests through https://github.com/canonical/data-platform-libs/pull/81?

paulomach: :thinking: Don't know about that. I've removed a test for the `_on_upgrade_granted` being abstract (which is not anymore)

marceloneppel: I merged main into this branch on https://github.com/canonical/data-platform-libs/pull/82/commits/cc95e4c46042281acd1a41376697d63b01ebb767.

marceloneppel (approved):

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -517,6 +542,18 @@ def upgrade_stack(self, stack: List[int]) -> None:
...
+    def unit_states(self) -> list:
+        """Current upgrade state for all units.
+
+        Returns:
+            Unsorted list of upgrade states for all units.
+        """
+        if not self.peer_relation:
+            return []
+
+        return [self.peer_relation.data[unit].get("state", "") for unit in self.app_units]
```
marcoppenheimer, line 555: **praise:** Useful!

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -692,6 +734,37 @@ def _on_pre_upgrade_check_action(self, event: ActionEvent) -> None:
...
+        if not self.upgrade_stack:
+            event.fail(message="Nothing to resume, upgrade stack unset.")
+            return
+
+        # Check whether this is being run after juju refresh was called
+        # (the size of the upgrade stack should match the number of total
+        # unit minus one).
+        if len(self.upgrade_stack) != len(self.peer_relation.units):
+            event.fail(message="Upgrade can be resumed only once after juju refresh is called.")
+            return
```
marcoppenheimer, line 759: **question:** Not having looked at it too deeply, what happens here when, say, we call this on the 4th out of 5 units?

The stack would be something like `[0, 1]`, but the number of peer units would be 5. Does that mean that the `resume-upgrade` fails here?

marcoppenheimer (approved):
Look good!

lib/charms/data_platform_libs/v0/upgrade.py
paulomach: Yes, as it should. The idea is that to be run only once, after which the rolling upgrade is triggered automatically at the end of a successful upgrade on each unit.

PR #84, opened 2023-08-10
[DPE-2394] Juju3 pipeline

juditnovak:
Juju 3 matrix pipelines for data-platform-libs.

**NOTE**: Due to Juju bug [Inconsistent state in EnterScope](https://bugs.launchpad.net/juju/+bug/2031631) a workaround is also applied, making pipeline execution a bit more reliable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
           - integration-opensearch
           - integration-kafka
         juju-version:
-          - juju-bootstrap-option: "2.9.38"
+          - juju-bootstrap-option: "2.9.44"
             juju-snap-channel: "2.9/stable"
             libjuju-version: "2.9.42.4"
           - juju-bootstrap-option: "3.1.5"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,14 @@ jobs:
           - integration-s3
           - integration-opensearch
           - integration-kafka
-    name: ${{ matrix.tox-environments }}
+        juju-version:
+          - juju-bootstrap-option: "3.1.5"
+            juju-snap-channel: "3.1/stable"
+            libjuju-version: "==3.2.0.1"
+          - juju-bootstrap-option: "2.9.38"
+            juju-snap-channel: "2.9/stable"
+            libjuju-version: "==2.9.42.4"
+    name: ${{ matrix.tox-environments }} Juju ${{ matrix.juju-version.juju-snap-channel}} -- ${{ matrix.juju-version.libjuju-version }}
     needs:
       - lint
       - unit-test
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -92,6 +108,9 @@ jobs:
         run: tox run -e ${{ matrix.tox-environments }} -- -m '${{ steps.select-tests.outputs.mark_expression }}' --model testing
         env:
           CI_PACKED_CHARMS: ${{ needs.build.outputs.charms }}
+          LIBJUJU_VERSION_SPECIFIER: "==${{ matrix.juju-version.libjuju-version }}"
+      - name: Print debug-log
+        run: juju switch testing; juju debug-log --replay --no-tail
```
marceloneppel, line 113: Can the dump logs step do this?

marceloneppel (approved):

.github/workflows/ci.yaml
juditnovak: Normally the `Dump logs` step is only dumping on failure. I looked into this earlier, but it seemed to me that the corresp. github action can't dump logs *without* a failure :-) 

Which is the reason why I went for this solution. 

- `Print debug-log` is *only* executed on success (skipped if there's a failure).
- While `Dump logs` is *only* executed on failure :-)

So we're getting the logs in both cases :-) 

However in case there may be a more professional solution here, I'm happy to take that onboard for sure.

.github/workflows/ci.yaml
```diff
@@ -56,7 +63,14 @@ jobs:
           - integration-s3
           - integration-opensearch
           - integration-kafka
-    name: ${{ matrix.tox-environments }}
+        juju-version:
+          - juju-bootstrap-option: "2.9.38"
```
paulomach, line 67: bump to current?
```suggestion
          - juju-bootstrap-option: "2.9.44"
```

tests/integration/database-charm/src/charm.py
```diff
@@ -100,6 +102,16 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
 
+        # Temporary hack to avoid https://bugs.launchpad.net/juju/+bug/2031631
+        sleep(randrange(3))
+
+        assert self.model.get_binding("database")
+        assert self.model.get_binding("database").network
+        assert self.model.get_binding("database").network.bind_address
+        logger.info(
+            f"Charm binding {self.model.get_binding('database')}, network: {self.model.get_binding('database').network}, IP:  self.model.get_binding('database').network.bind_address"
+        )
```
paulomach, line 113: fmt nit:
```suggestion
        logger.info(
            (
                f"Charm binding {self.model.get_binding('database')}, "
                f"network: {self.model.get_binding('database').network}, "
                f"IP: {self.model.get_binding('database').network.bind_address}"
            )
        )
```

paulomach (approved):
lgtm, small non-blockers

marceloneppel (approved):

paulomach (approved):
exquisite

PR #87, opened 2023-08-17
fix: upgrade idle state

zmraul:
Fixes the `idle` property so the result is the expected from the docstring.


marcoppenheimer (approved):

marceloneppel (approved):
LGTM!

paulomach (approved):

PR #88, opened 2023-08-22
clean exit when a unit not yet in the upgrade relation

paulomach:
### Issue

**edge case**: When refreshing from a charm version without upgrade support, `on_upgrade_changed` hook error if the top_stack unit has not yet joined the relation.

**context**: mysql need to be able to support upgrade from the current stable version (if possible) to minimize noise from users. Check [PR301](https://github.com/canonical/mysql-operator/pull/301)

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -284,7 +284,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 PYDEPS = ["pydantic>=1.10,<2"]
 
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -1127,7 +1127,14 @@ def on_upgrade_changed(self, event: EventBase) -> None:
...
         top_unit = self.charm.model.get_unit(f"{self.charm.app.name}/{top_unit_id}")
-        top_state = self.peer_relation.data[top_unit].get("state")
+        try:
+            top_state = self.peer_relation.data[top_unit].get("state")
+        except KeyError:
+            # a unit may not have joined the relation when  upgrading from
+            # an older version of the charm without upgrade support
+            # the hook will retrigger when it joins
+            logger.debug(f"{top_unit} not found in peer relation data, exiting...")
+            return
```
zmraul, line 1137: Is this a race condition? During the whole function, there are several calls to `self.cluster_status`, `self.unit_states`... all of which invoke `self.peer_relation.data[unit].get("state")` for all units.

For this error to trigger, it would mean that the relation exists at the beginning of the function, and the unit ID has beed added to the stack. But at some point in between L1124 and L1131 the unit dissapears?

marcoppenheimer:
@paulomach - AFAIK, we weren't supporting upgrades for versions 'before' implementation of `build_upgrade_stack`. 

As (I'm assuming) we **have** to support it, aside from this PR, how are we getting the stack in to relation data 'manually'?

paulomach:
@zmraul - yes, it's related to workaround to make upgrade work on non-upgrade enabled charms. And I can workaround on the charm side - closing this.
@marcoppenheimer - we want to mitigate noise from current users of the stable channel.

PR #89, opened 2023-08-22
[MISC] Improve optional types support

deusebio:


tests/unit/test_data_models.py
```diff
@@ -201,3 +209,64 @@ def test_relation_databag_merged(self):
...
+    @parameterized.expand(
+        [
+            ("option-float", "1.0", float),
+            ("option-float", "1", float),
+            ("option-int", "1", int),
+            ("option-str", "1", str),
+            ("option-str", "test", str),
+        ]
+    )
+    def test_relation_databag_merged_with_option(self, option_key, option_value, _type):
```
welpaolo, line 225: Can you please add the docstring to new tests ?

welpaolo (approved):
Looks good to me! 
Just left a really minor comment!

tests/unit/test_data_models.py
deusebio: Actually also old tests did not have doc strings :O I have updated them all in the data_models module

welpaolo: Nice! Thank you so much

tests/unit/test_data_models.py
```diff
@@ -201,3 +215,68 @@ def test_relation_databag_merged(self):
...
+
+    @parameterized.expand(
+        [
+            ("option-float", "1.0", float),
+            ("option-float", "1", float),
+            ("option-int", "1", int),
+            ("option-str", "1", str),
+            ("option-str", "test", str),
+        ]
+    )
```
marcoppenheimer, line 230: **question:** Why not `pytest.mark.parameterize`?
If there wasn't a main reason, I would recommend we use that.

marcoppenheimer: Oh, it's `unittest`, right?

marcoppenheimer (approved):

tests/unit/test_data_models.py
deusebio: yes, that's right. Actually, some tests used `unittests`, some used `pytest`. We should probably unify this at a certain point

PR #90, opened 2023-08-23
Removed unnedeed defer

paulomach:
Event defer is not needed.


Mehdi-Bendriss (approved):

welpaolo (approved):
I will suggest to update the log message above!

PR #91, opened 2023-08-26
[DPE-2342] Charm Relation (i.e. 'external') Secrets

juditnovak:
## Disclaimer

Here we introduce Relations Secrets on top of the existing `data_interfaces` logic.

New interfaces:
   - `get_relation_secret_data()` (`Provides`, `Requires`)
   - `set_relation_secret_fields()` (`Provides`)
   - `get_relation_secret_fields()`, `get_relation_secret_fields()` (`Requires`)

POC demo of usage demonstrated on Opensearch: https://github.com/canonical/opensearch-operator/pull/116

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -296,8 +296,10 @@ import logging
 from abc import ABC, abstractmethod
 from collections import namedtuple
 from datetime import datetime
-from typing import List, Optional, Union
+from enum import Enum
+from typing import Dict, List, Optional, Union
 
+from ops import Handle, JujuVersion, Secret
 from ops.charm import (
     CharmBase,
     CharmEvents,
@@ -312,11 +314,11 @@ from ops.model import Application, ModelError, Relation, Unit
 LIBID = "6c3e6b6680d64e9c89e611d1a15f65be"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 0
+LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 0
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -329,6 +331,18 @@ A tuple for storing the diff between two data mappings.
 added - keys that were added
 changed - keys that still exist but have new values
 deleted - key that were deleted"""
+
+
+class DataInterfacesError(Exception):
+    """Common ancestor for Secrets related exceptions."""
+
+
+class SecretError(Exception):
+    """Common ancestor for Secrets related exceptions."""
+
+
+class SecretAlreadyExistsError(SecretError):
+    """A secret that was to be added already exists."""
 
 
 def diff(event: RelationChangedEvent, bucket: Union[Unit, Application]) -> Diff:
@@ -365,6 +379,81 @@ def diff(event: RelationChangedEvent, bucket: Union[Unit, Application]) -> Diff:
     return Diff(added, changed, deleted)
 
 
+class Scope(Enum):
+    """Peer relations scope."""
+
+    APP = "app"
+    UNIT = "unit"
+
+
+class SecretCache:
+    """Internal helper class locally cache secrets.
+
+    The data structure is precisely re-using/simulating as in the actual Secret Storage
+    """
+
+    def __init__(self, charm: CharmBase, secret_uri: Optional[str] = None):
+        # Structure:
+        # NOTE: "objects" (i.e. dict-s) and scalar values are handled in a unified way
+        # precisely as done for the Secret objects themselves.
+        #
+        # self.secrets = {
+        #   "app": {
+        #       "opensearch:app:admin-password": {
+        #           "meta": <Secret instance>,
+        #           "content": {
+        #               "opensearch:app:admin-password": "bla"
+        #           }
+        #       }
+        #   },
+        #   "unit": {
+        #       "opensearch:unit:0:certificates": {
+        #           "meta": <Secret instance>,
+        #           "content": {
+        #               "ca-cert": "<certificate>",
+        #               "cert": "<certificate>",
+        #               "chain": "<certificate>"
+        #           }
+        #       }
+        #   }
+        # }
+        self._secret_meta = None
+        self._secret_content = {}
+        self._secret_uri = secret_uri
+        self.charm = charm
+
+    def add_secret(self, content: Dict[str, str], relation: Relation) -> Secret:
+        """Create a new secret."""
+        if self._secret_uri:
+            raise SecretAlreadyExistsError("Secret is already defined for")
+        secret = self.charm.app.add_secret(content)
+        secret.grant(relation)
+        self._secret_uri = secret.id
+        self._secret = secret
+        return self._secret
+
+    def get_secret(self) -> Optional[Secret]:
+        """Getting cached secret meta-information."""
+        if not self._secret_meta:
+            if not self._secret_uri:
+                return
+            self._secret_meta = self.charm.model.get_secret(id=self._secret_uri)
+        return self._secret_meta
+
+    def get_content(self, scope: Scope, label: str) -> Dict[str, str]:
+        """Getting cached secret content."""
+        if not self._secret_content:
+            if secret_meta := self.get_secret():
+                self._secret_content = secret_meta.get_content()
+        return self._secret_content
+
+    def set_content(self, scope: Scope, label: str, content: Dict[str, str]) -> None:
+        """Setting cached secret content."""
+        if secret_meta := self.get_secret():
+            secret_meta.set_content(content)
+            self._secret_content = content
+
+
 # Base DataRelation
 
 
@@ -381,11 +470,75 @@ class DataRelation(Object, ABC):
             charm.on[relation_name].relation_changed,
             self._on_relation_changed_event,
         )
+        self._jujuversion = None
+        self.secrets = {}
+
+    @property
+    def relations(self) -> List[Relation]:
+        """The list of Relation instances associated with this relation_name."""
+        return [
+            relation
+            for relation in self.charm.model.relations[self.relation_name]
+            if self._is_relation_active(relation)
+        ]
+
+    @property
+    def secrets_enabled(self):
+        """Is this Juju version allowing for Secrets usage?"""
+        if not self._jujuversion:
+            self._jujuversion = JujuVersion.from_environ()
+        return self._jujuversion.has_secrets
 
     @abstractmethod
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the relation data has changed."""
         raise NotImplementedError
+
+    @staticmethod
+    def _is_relation_active(relation: Relation):
+        """Whether the relation is active based on contained data."""
+        try:
+            _ = repr(relation.data)
+            return True
+        except (RuntimeError, ModelError):
+            return False
+
+    def get_relation(self, relation_name, relation_id) -> Relation:
+        """Safe way of retrieving a relation."""
+        relation = self.charm.model.get_relation(relation_name, relation_id)
+
+        if not relation:
+            raise DataInterfacesError(
+                "Relation %s %s couldn't be retrieved", relation_name, relation_id
+            )
+
+        if not relation.app:
+            raise DataInterfacesError("Relation's application missing")
+
+        return relation
+
+    def get_relation_secret(
+        self, relation_id: int, relation_name: Optional[str] = None
+    ) -> Optional[Secret]:
+        """Retrieve a Juju Secret that's been stored in the relation databag."""
+        if not relation_name:
+            relation_name = self.relation_name
+
+        if not self.secrets.get(relation_id) and self.secrets_enabled:
+            relation = self.charm.model.get_relation(relation_name, relation_id)
+            if not relation:
+                return
+            if secret_id := relation.data[self.local_app].get("secret"):
+                self.secrets[relation_id] = SecretCache(self.charm, secret_id).get_secret()
+        return self.secrets.get(relation_id)
+
+    def get_relation_secret_data(
+        self, relation_id: int, relation_name: Optional[str] = None
+    ) -> Optional[Dict[str, str]]:
+        """Retrieve contents of a Juju Secret that's been stored in the relation databag."""
+        secret = self.get_relation_secret(relation_id, relation_name)
+        if secret:
+            return secret.get_content()
 
     def fetch_relation_data(self) -> dict:
         """Retrieves data from relation.
@@ -423,24 +576,6 @@ class DataRelation(Object, ABC):
             if relation:
                 relation.data[self.local_app].update(data)
 
-    @staticmethod
-    def _is_relation_active(relation: Relation):
-        """Whether the relation is active based on contained data."""
-        try:
-            _ = repr(relation.data)
-            return True
-        except (RuntimeError, ModelError):
-            return False
-
-    @property
-    def relations(self) -> List[Relation]:
-        """The list of Relation instances associated with this relation_name."""
-        return [
-            relation
-            for relation in self.charm.model.relations[self.relation_name]
-            if self._is_relation_active(relation)
-        ]
-
 
 # Base DataProvides and DataRequires
 
@@ -463,6 +598,66 @@ class DataProvides(DataRelation):
         """
         return diff(event, self.local_app)
 
+    def add_relation_secret(self, relation_id, content) -> Optional[Secret]:
+        """Add a new Juju Secret that will be registered in the relation databag."""
+        if not self.local_unit.is_leader():
+            return
+
+        relation = self.get_relation(self.relation_name, relation_id)
+
+        if relation.data[self.local_app].get("secret"):
+            logging.error("Secret for relation %s already exists, not adding again", relation_id)
+            return
+
+        secret = SecretCache(self.charm).add_secret(content, relation)
+
+        # According to lint we may not have a Secret ID
+        if secret.id:
+            relation.data[self.local_app]["secret"] = secret.id
+            self.secrets[relation_id] = secret
+            return secret
+
+    def update_relation_secret(self, relation_id, content):
+        """Update the contenst of an existing Juju Secret, stored in the relation databag."""
+        if not self.local_unit.is_leader():
+            return
+
+        secret = self.get_relation_secret(relation_id)
+
+        if not secret:
+            logging.error("Can't update secret for relation %s", relation_id)
+            return
+
+        full_content = secret.get_content()
+        full_content.update(content)
+        secret.set_content(full_content)
+
+    def set_relation_fields(self, relation_id: int, fields: Dict[str, str]) -> None:
+        """Get the value of a field not caring whether it's a secret or not."""
+        relation = self.get_relation(self.relation_name, relation_id)
+        relation_secret_fields = relation.data.get(
+            relation.app, {}  # pyright: ignore [reportGeneralTypeIssues]
+        ).get("secret_fields")
+
+        normal_fields = list(fields)
+        if relation_secret_fields:
+            relation_secret_fields = relation_secret_fields.split(" ")
+            normal_fields = set(fields.keys()) - set(relation_secret_fields)
+            secret_fields = set(fields.keys()) - set(normal_fields)
+
+            if secret_fields and self.secrets_enabled:
+                secret_content = {k: v for k, v in fields.items() if k in secret_fields}
+
+                if self.get_relation_secret(relation_id):
+                    self.update_relation_secret(relation_id, secret_content)
+                else:
+                    self.add_relation_secret(relation_id, secret_content)
+
+        normal_content = {k: v for k, v in fields.items() if k in normal_fields}
+        relation.data[self.local_app].update(  # pyright: ignore [reportGeneralTypeIssues]
+            normal_content
+        )
+
     def set_credentials(self, relation_id: int, username: str, password: str) -> None:
         """Set credentials.
 
@@ -474,13 +669,7 @@ class DataProvides(DataRelation):
             username: user that was created.
             password: password of the created user.
         """
-        self._update_relation_data(
-            relation_id,
-            {
-                "username": username,
-                "password": password,
-            },
-        )
+        self.set_relation_fields(relation_id, {"username": username, "password": password})
 
     def set_tls(self, relation_id: int, tls: str) -> None:
         """Set whether TLS is enabled.
@@ -489,7 +678,7 @@ class DataProvides(DataRelation):
             relation_id: the identifier for a particular relation.
             tls: whether tls is enabled (True or False).
         """
-        self._update_relation_data(relation_id, {"tls": tls})
+        self.set_relation_fields(relation_id, {"tls": tls})
 
     def set_tls_ca(self, relation_id: int, tls_ca: str) -> None:
         """Set the TLS CA in the application relation databag.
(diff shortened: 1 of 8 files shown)


tests/unit/test_data_interfaces.py
```diff
@@ -250,6 +283,37 @@ def test_set_additional_fields(self):
...
+        assert secret_id
+
+        relation_data == {
+            "data": '{"secret_fields": "'
+            + f"{self.SECRET_FIELDS}"
+            + '"}',  # Data is the diff stored between multiple relation changed events.   # noqa
+            "replset": "rs0",
+            "version": "1.0",
+            "uris": "host1:port,host2:port",
+        }
```
marceloneppel, line 308: Is it missing to check if this data is in the relation? The same for `test_set_additional_fields_secrets` and `test_set_additional_fields_secrets`.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -498,12 +687,14 @@ def set_tls_ca(self, relation_id: int, tls_ca: str) -> None:
...
             tls_ca: TLS certification authority.
         """
-        self._update_relation_data(relation_id, {"tls-ca": tls_ca})
+        self.set_relation_fields(relation_id, {"tls-ca": tls_ca})
 
 
 class DataRequires(DataRelation):
     """Requires-side of the relation."""
 
+    SECRET_FIELDS = ["username", "password", "tls", "tls_ca"]
```
marceloneppel, line 696: Is it extendable to a list of additional fields that the requirer would like to be protected through secrets, like we wrote in the spec?

tests/integration/kafka-charm/src/charm.py
```diff
@@ -108,7 +108,13 @@ def _on_topic_requested(self, event: TopicRequestedEvent):
     def _on_sync_password(self, event: ActionEvent):
         """Set the password in the data relation databag."""
         logger.info("On sync password")
-        password = event.params["password"]
+
+        password = event.params.get("password")
+        if not password:
+            secret_uri = event.params.get("secret")
+            secret_contents = self.charm.model.get_secret(id=secret_uri).get_content()
+            password = secret_contents.get("password")
```
marceloneppel, line 116: Is it possible to use `get_relation_secret_data` here and below? It will be good to have it as an example for charmers on how to implement the secrets retrieval using the right method (and use the cache).

marceloneppel (commented):
Overall it looks great! I think we are in the right path with this implementation.

About the cache, from what I could see, it will be restricted to the event lifetime only.

I left comments about other details.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Yes, this was one of the missing items on this Draft :-)
I'm adding them now that the base concept is confirmed :-)

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -365,6 +379,81 @@ def diff(event: RelationChangedEvent, bucket: Union[Unit, Application]) -> Diff:
...
+        self.charm = charm
+
+    def add_secret(self, content: Dict[str, str], relation: Relation) -> Secret:
+        """Create a new secret."""
+        if self._secret_uri:
+            raise SecretAlreadyExistsError("Secret is already defined for")
+        secret = self.charm.app.add_secret(content)
+        secret.grant(relation)
+        self._secret_uri = secret.id
+        self._secret = secret
```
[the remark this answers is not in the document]
juditnovak: VERY GOOD CALL, thank U so much....

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -534,14 +727,44 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+        self, relation_id: int, fields: List[str], relation_name: Optional[str] = None
+    ) -> Dict[str, str]:
+        """Get the value of a field not caring whether it's a secret or not."""
+        if not relation_name:
+            relation_name = self.relation_name
+
+        result = {}
+        relation = self.get_relation(relation_name, relation_id)
+        relation_secret_fields = relation.data.get(
+            relation.app, {}  # pyright: ignore [reportGeneralTypeIssues]
```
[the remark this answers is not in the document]
juditnovak: Good point, thank you!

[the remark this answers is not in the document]
juditnovak: Using `self.local_app` for consistency (and I guess to follow design guidelines).
It's pointing to `self.charm.model.app` but I assume it's the same :-)

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1013,7 +1250,12 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
 
         # Check if the database is created
         # (the database charm shared the credentials).
-        if "username" in diff.added and "password" in diff.added:
+
+        new_secret_content = None
+        if self.secrets_enabled and "secret" in diff.added:
+            new_secret_content = self.get_relation_fields(event.relation.id, ["username", "password"])
+
+        if ("username" in diff.added and "password" in diff.added) or new_secret_content:
```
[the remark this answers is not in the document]
juditnovak: Sadly that's not quite the way it goes with secrets. I close this discussion and we should get back to this offline.

It's a lot more complicated than that -- thus I've implemented a solution based on revisions, which seems to provide at least a reasonable workaround for now...

tests/integration/kafka-charm/src/charm.py
juditnovak: Totally, good point!

tests/unit/test_data_interfaces.py
marceloneppel: My bad, I didn't see that it was a comparison instead of an assignment.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +588,104 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+
+        relation = self.get_relation(self.relation_name, relation_id)
+
+        if relation.data[self.local_app].get(f"secret-{label}"):
+            logging.error("Secret for relation %s already exists, not adding again", relation_id)
+            return
+
+        secret = SecretCache(self.charm).add_secret(content, relation)
+
+        # According to lint we may not have a Secret ID
```
shayancanonical, line 606: may be we should add a lint exception? also, how should we be handling exceptions when calling `add_secret`? if we dont handle it, the database charm will go into error state and retry the handler that wanted to set relation data

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -534,14 +755,65 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+            secret_fields = set(fields) - set(normal_fields)
+
+            if secret_fields and self.secrets_enabled:
+                label_sorted_content = self._create_label_sorted_content(fields)
+                for label in label_sorted_content:
+                    if secret := self.get_relation_secret(relation_id, label):
+                        if secret_data := secret.get_content():
+                            result.update(
+                                {k: v for k, v in secret_data.items() if k in secret_fields}
+                            )
```
shayancanonical, line 801: stylistic nit: i feel like the cyclomatic complexity of this code block is too high -> suggest trying to move block 794-801 outside of the if statement on line 789

```suggestion
        normal_fields, secret_fields = fields, []
        if relation_secret_fields:
            relation_secret_fields = relation_secret_fields.split(" ")
            normal_fields = set(fields) - set(relation_secret_fields)
            secret_fields = set(fields) - set(normal_fields)

        if secret_fields and self.secrets_enabled:
            label_sorted_content = self._create_label_sorted_content(fields)
            for label in label_sorted_content:
                if (secret := self.get_relation_secret(relation_id, label)) and (secret_data := secret.get_content()(:
                    result.update(
                        {k: v for k, v in secret_data.items() if k in secret_fields}
                    )
```

shayancanonical (commented):
I really like the direction that this implementation is taking! Abstracting away whether a relation data is stored in a secrets backend or in the relational databag is an excellent idea!

A couple of high level comments/questions I had after my first review were:
1. SecretCache is not really a cache - it is a wrapper around a Secret in a Juju secrets backend
2. Should we allow the Requires application to decide which fields are `SECRET_FIELDS` instead of hardcoding them? (and fallback to the hardcoded value if they are not specified?) -> this will allow users to specify all the fields that they want as secrets

juditnovak:
@shayancanonical 

SecretCache: Your point is valid, it is 

 1. An abstraction layer to handle direct interaction with the Juju Secret Strore
 2. **However** it IS a cache too :-)
 
 The point is that, once you fetch a secret via a SecretCache object, it won't be fetched again within the same event scope.
 So it's kinda providing a smart layer between DataInterfaces and Juju. It's both handling the interactions AND if the interaction is avoidable it would just return the previously cached data.
 
 (I may need to polish up a code, there could be errors on this still :-) )

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +598,66 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+
+                if self.get_relation_secret(relation_id):
+                    self.update_relation_secret(relation_id, secret_content)
+                else:
+                    self.add_relation_secret(relation_id, secret_content)
+
+        normal_content = {k: v for k, v in fields.items() if k in normal_fields}
+        relation.data[self.local_app].update(  # pyright: ignore [reportGeneralTypeIssues]
+            normal_content
+        )
```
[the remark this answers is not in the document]
juditnovak: Fixed, and checked for similar occurance.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: So, to reflect offline conversation here:We should be throwing exceptions on critical error (typically, secret couldn't be created/updated/fetched). The code is doing that now.

marceloneppel (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -418,28 +588,16 @@ def _update_relation_data(self, relation_id: int, data: dict) -> None:
...
-
-    @property
-    def relations(self) -> List[Relation]:
-        """The list of Relation instances associated with this relation_name."""
-        return [
-            relation
-            for relation in self.charm.model.relations[self.relation_name]
-            if self._is_relation_active(relation)
-        ]
+    def _create_label_sorted_content(keys: List[str]) -> Dict[str, List[str]]:
```
shayancanonical, line 596: minor nit: missing docstring

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +621,102 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+        # According to lint we may not have a Secret ID
+        if secret.meta and secret.meta.id and (secret_info := secret.get_info()):
+            relation.data[self.local_app][f"secret-{label}"] = secret.meta.id
+            relation.data[self.local_app][f"secret-{label}-revision"] = str(secret_info.revision)
+            self.secrets.setdefault(relation_id, {})[label] = secret
+
+    @leader_only
+    @juju_secrets_only
+    def update_relation_secret(self, relation_id: int, content: Dict[str, str], label: str):
+        """Update the contenst of an existing Juju Secret, referred in the relation databag."""
```
shayancanonical, line 648: minor nit: contenst -> contents

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +621,102 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+        secret = self._get_relation_secret(relation_id, label)
+
+        if not secret:
+            logging.error("Can't update secret for relation %s", relation_id)
+            return
+
+        old_content = secret.get_content()
+        full_content = copy.deepcopy(old_content)
+        full_content.update(content)
```
shayancanonical, line 657: do we want to only update the keys in `content` or should we be updating the content of the secret with `content`?

if old_content is `{'a': 'A'}` and content is `{'b': 'B'}`, we will the secret's content to `{'a': 'A', 'b': 'B'}`
if this is desired behavior, we should note this in the method's docstring -> that it is not possible (or even needed) to remove key/value pairs from a secret

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +621,102 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+
+    @juju_secrets_only
+    def _get_relation_secret(
+        self, relation_id: int, label: str, relation_name: Optional[str] = None
+    ) -> Optional[SecretCache]:
+        """Retrieve a Juju Secret that's been stored in the relation databag."""
+        if not relation_name:
+            relation_name = self.relation_name
+
+        if not self.secrets.get(relation_id, {}).get(label) and self.secrets_enabled:
```
shayancanonical, line 677: do we need to check `self.secrets_enabled` (since we are using the @juju_secrets_only decorator)?

also, very minor nit (feel free to ignore): i prefer the following structure with early-outs as it is easier imo to track the code logic:

```
secret = self.secrets.get(relation_id, {}).get(label)
if secret:
    return secret

relation = self.charm.model.get_relation(relation_name, relation_id)
if not relation:
    return None

if secret_id := relation.data[self.local_app].get(f"secret-{label}"):
    self.secrets.setdefault(relation_id, {})[label] = SecretCache(self.charm, secret_id)

return self.secrets.get(relation_id, {}).get(label)
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -463,6 +621,102 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
+                return
+            if secret_id := relation.data[self.local_app].get(f"secret-{label}"):
+                self.secrets.setdefault(relation_id, {})[label] = SecretCache(
+                    self.charm, secret_id
+                )
+        return self.secrets.get(relation_id, {}).get(label)
+
+    @leader_only
+    def set_relation_fields(self, relation_id: int, fields: Dict[str, str]) -> None:
+        """Get the value of a field not caring whether it's a secret or not."""
```
shayancanonical, line 689: Not sure that the docstring here is accurate

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -517,10 +767,19 @@ def __init__(
...
+    @property
+    def secret_fields(self) -> Optional[List[str]]:
+        """Local access to secrets field, in case they are being used."""
+        if self.secrets_enabled:
+            return self.SECRET_FIELDS
+
     def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
         """Event emitted when the relation is created."""
-        raise NotImplementedError
+        if not self.local_unit.is_leader():
```
shayancanonical, line 778: can we use the @leader_only decorator here?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -761,10 +1077,9 @@ def __init__(self, charm: CharmBase, relation_name: str) -> None:
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the relation has changed."""
-        # Only the leader should handle this event.
+        # Leader only
```
shayancanonical, line 1080: can we use the @leader_only decorator here too?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -534,14 +793,66 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
...
-    def _is_resource_created_for_relation(relation: Relation) -> bool:
+    @juju_secrets_only
+    def _get_relation_secret(
+        self, relation_id: int, label: str, relation_name: Optional[str] = None
+    ) -> Optional[SecretCache]:
+        """Retrieve a Juju Secret that's been stored in the relation databag."""
+        if not relation_name:
+            relation_name = self.relation_name
+
+        if not self.secrets.get(relation_id, {}).get(label) and self.secrets_enabled:
```
shayancanonical, line 804: do we need self.secrets_enabled here (due to the @juju_secrets_only decorator)?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1159,7 +1485,7 @@ def __init__(self, charm: CharmBase, relation_name: str) -> None:
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the relation has changed."""
-        # Only the leader should handle this event.
+        # Leader only
```
shayancanonical, line 1488: can we use the @leader_only decorator?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1339,10 +1673,9 @@ def __init__(self, charm: CharmBase, relation_name: str) -> None:
...
```
shayancanonical, line 1676: can we use the @leader_only decorator?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1415,7 +1750,7 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         diff = self._diff(event)
 
         # Check if authentication has updated, emit event if so
-        updates = {"username", "password", "tls", "tls-ca"}
+        updates = {"username", "password", "tls", "tls-ca", "secret-user"}
```
shayancanonical, line 1753: do we need to add `"secret-tls"` to updates?

shayancanonical (commented):
Looks great! Just a flood of small nits/questions

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: No, not on event handlers :-(

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Great point thank you!!!!!!!!!!!!!!!!!!!!!

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: No, not on event handelers :-(

I don't have the exact error message anymore, I think it was on charmcraft level. Simply event handlers can't have decorators, it messes up something with parameters passing

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: No, not on an event handler :-(

(Dupliate closing.)

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Duplicate, closing.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: No, not on event handlers.

(Duplicate, closing)

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: So far as I understood the logic, there is no need for deletion. Instead we are either adding to secrets (initial content) or we're modifying existing content. So far there's no action using a deletion. I think if it would be needed, additional extension may be necessary -- in a separate PR :grin:

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Fair point on `secrets_enabled`. It is redundant once the decorator is used.

The property still has its value, as it is to support decision-making on workflows that are to be executable on both Juju versions.

I'll remove it from redundant calls, thank you :-)

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Duplicate of https://github.com/canonical/data-platform-libs/pull/91#discussion_r1310821963

paulomach (approved):
I'm missing some of the secret context, but could follow the implementation. Overall looking good, my nits are well represented by other reviews

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -595,6 +907,11 @@ def extra_user_roles(self) -> Optional[str]:
 class AuthenticationEvent(RelationEvent):
     """Base class for authentication fields for events."""
 
+    def __init__(
+        self, handle: Handle, relation: Relation, app: Optional[Application], unit: Optional[Unit]
+    ):
+        super().__init__(handle, relation, app, unit)
+
     @property
```
juditnovak, line 915: NOTE: The fields below are to be removed [in the v0 -> <- v1 PR](https://github.com/canonical/data-platform-libs/pull/92).

If you have comments on that, pls do it on this thread

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -595,38 +909,6 @@ def extra_user_roles(self) -> Optional[str]:
 class AuthenticationEvent(RelationEvent):
     """Base class for authentication fields for events."""
```
juditnovak, line 911: ACTUAL BREAKING CHANGE

shayancanonical (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -593,14 +905,61 @@ def extra_user_roles(self) -> Optional[str]:
...
+        return self._cached_secrets
+
+    @property
+    def _jujuversion(self) -> JujuVersion:
+        """Caching jujuversion to avoid a Juju call on each field evaluation.
+
+        DON'T USE the encapsulated helper variable outside of this function
+        """
+        if not hasattr(self, "_cached_jujuversion"):
+            self._cached_jujuversion = None
```
juditnovak, line 931: Alternatively, I can add this __init__() function:

```
def __init__(self, handle: Handle, relation: Relation, app: Optional[Application] = None, unit: Optional[Unit] = None):
    super().__init__(handle, relation, app, unit)
    self._cached_jujuversion = None
    self._cached_secrets = None
````
...and then copy this into 6 --currently very pretty, one-liner-- descendant classes:
```
def __init__(self, handle: Handle, relation: Relation, app: Optional[Application] = None, unit: Optional[Unit] = None):
    super().__init__(handle, relation, app, unit)
```
...but this is so ugly, that it doesn't even feel right -- just for two internal helper variables, that aren't supposed to be used outside of their encapsulated scope.... That I'd rather go with the technically less safe, but in another sense still more elegant solution. As instructions on the property functions clearly indicate: the helper variabes are meant to be encapsulated inside, noone should refer to them otherwise.

Anyway, if strong feelings on the other side, I can go for the `__init__()` solution.

Comment (paulomach) [reply lib/charms/data_platform_libs/v0/data_interfaces.py:931]: less ugly?

```suggestion
        if not hasattr(self, "_cached_jujuversion") or not self._cached_jujuversion:
            self._cached_jujuversion = JujuVersion.from_environ()
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -331,6 +334,33 @@ def _on_topic_requested(self, event: TopicRequestedEvent):
 deleted - key that were deleted"""
 
 
+# Local map to associate labels with secrets potentially as a group
```
juditnovak, line 337: [CRITICAL] Increase `LIBVERSION` !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

shayancanonical:
@juditnovak is this PR still revelant? if not, should we close?

PR #93, opened 2023-09-06
Use poetry-core to parse version constraints

carlcsaposs-canonical:


marcoppenheimer:
Such a good addition, though it breaks my heart that it was that easy.

taurus-forever (commented):
Up 2 Marc here... P.S. massive codebase simplification. Nice!
Can we keep tests untouched to guaranties full compatibility?

marcoppenheimer (approved):

taurus-forever (approved):

PR #95, opened 2023-09-18
Fix off-by-1 error in upgrade.py

carlcsaposs-canonical:
```python
>>> len([0, 1, 2])
3
```

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -1019,7 +1019,7 @@ def _on_upgrade_finished(self, _) -> None:
         # This hook shouldn't run for the last unit (the first that is upgraded). For that unit it
         # should be done through an action after the upgrade success on that unit is double-checked.
         unit_number = int(self.charm.unit.name.split("/")[1])
-        if unit_number == len(self.peer_relation.units):
+        if unit_number == len(self.peer_relation.units) - 1:
```
marceloneppel, line 1022: AFAIK, in a 3 units cluster we should have 2 units in the peer relation object (the other units, without the current unit).

marcoppenheimer: I agree it's a bit obtuse.
Maybe add a comment there, given that it's tripped someone up?

marceloneppel: A comment is a good idea.

marcoppenheimer: I think whether or not we want to support upgrades for 'new' peer-relations, is up to the charm author setting `upgrade_supported`. 
Personally I wouldn't want to use `planned_units`, as it's a bit indirect. There's value in using the units that actually are 'peered', because the whole upgrade depends on peer-relation data being exchanged.

taurus-forever: @marceloneppel @carlcsaposs-canonical @marcoppenheimer  please schedule a call to discuss this situation and clarify this case. I cannot explain why this was not noticed on upgrade testing. Are we testing/supporting single unit `juju refresh`? :-)

marcoppenheimer: @taurus-forever Just a point here, if it's single unit, I believe that all of our 'upgrade one by one' logic goes out the window, as we can't maintain service uptime if the only live unit goes down.

marcoppenheimer: Correct, by default. I believe that @paulomach worked around this for MySQL, is that right?

taurus-forever: > tl;dr; the code is correct

@carlcsaposs-canonical abandon PR?

PR #98, opened 2023-09-22
DPE-2660 VM specific defer

### LINKED ISSUE #97: For upgrade rollback on k8s, the `upgrade-relation-changed` handler will defer indefinitely
On k8s, for a rollback case, the leader unit will have the `recovery` state, rendering the `cluster_state` with the `recovery` value.

This will defer the `on_upgrade_changed` handler indefinitely because of [this test](https://github.com/canonical/data-platform-libs/blob/0275ffea7dbc3745ed1aa0bae1f2d9b364be985c/lib/charms/data_platform_libs/v0/upgrade.py#L924).

The defer makes sense for VM cases, where on charm refresh, all units will set it state to `ready` on the `upgrade-charm` handler.
On K8s, only the upgrading unit will receive the `upgrade-charm` event, and if the leader unit is not the upgrading unit, the `cluster_state` will remain in the recovery state.

Deferring the `on_upgrade_changed` indefinitely will prevent the `upgrade_stack` being popped, and a subsequent `resume-upgrade` will set the wrong partition value, staling the rollback process.

paulomach:
Address issue #97 
* defer for vm only
* fix log message

Fix #97

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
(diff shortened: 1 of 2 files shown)


zmraul (approved):
Neat

paulomach:
> Approving as it looks fine!
> 
> How are we doing rollbacks in K8s?

With this change, not different from VM, but re-rolling all pods. 
Carl is experimenting with a different approach (see [DA058](https://docs.google.com/document/d/1tLjknwHudjcHs42nzPVBNkHs98XxAOT2BXGGpP7NyEU/edit?usp=drivesdk))

PR #99, opened 2023-10-07
[DPE-2723] fetch_my_relation() data function

juditnovak:
Currently no interface access is provided to retrieve data for the relation's "owner" (i.e. "this app") side.

This proposal is to fill that gap

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -378,7 +378,7 @@ class SecretsIllegalUpdateError(SecretError):
 
 
 def get_encoded_field(
-    relation: Relation, member: Union[Unit, Application], field
+    relation: Relation, member: Union[Unit, Application], field: str
 ) -> Union[str, List[str], Dict[str, str]]:
     """Retrieve and decode an encoded field from relation data."""
     return json.loads(relation.data[member].get(field, "{}"))
@@ -408,9 +408,12 @@ def diff(event: RelationChangedEvent, bucket: Union[Unit, Application]) -> Diff:
     # Retrieve the old data from the data key in the application relation databag.
     old_data = get_encoded_field(event.relation, bucket, "data")
 
-    if not old_data or (old_data and not isinstance(old_data, dict)):
+    if not old_data:
+        old_data = {}
+
+    if not isinstance(old_data, dict):
         # We should never get here, added to re-assure pyright
-        logger.error("Can't retrieve previous databag diff.")
+        logger.error("Previous databag diff is of a wrong type.")
         old_data = {}
 
     # Retrieve the new data from the event relation databag.
@@ -604,14 +607,14 @@ class DataRelation(Object, ABC):
 
     @abstractmethod
     def _fetch_specific_relation_data(
-        self, relation, fields: Optional[List[str]]
+        self, relation: Relation, fields: Optional[List[str]]
     ) -> Dict[str, str]:
         """Fetch data available (directily or indirectly -- i.e. secrets) from the relation."""
         raise NotImplementedError
 
     @abstractmethod
     def _fetch_my_specific_relation_data(
-        self, relation, fields: Optional[List[str]]
+        self, relation: Relation, fields: Optional[List[str]]
     ) -> Dict[str, str]:
         """Fetch data available (directily or indirectly -- i.e. secrets) from the relation for owner/this_app."""
         raise NotImplementedError
@@ -684,7 +687,7 @@ class DataRelation(Object, ABC):
 
     def _retrieve_group_secret_contents(
         self,
-        relation_id,
+        relation_id: int,
         group: SecretGroup,
         secret_fields: Optional[Union[Set[str], List[str]]] = None,
     ) -> Dict[str, str]:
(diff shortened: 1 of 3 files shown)


juditnovak:
TODO : Increase LIBPATCH.
(Actually it's already dont locally, but I didn't wanna ruin the green pipelines :-) )

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -400,6 +407,12 @@ def diff(event: RelationChangedEvent, bucket: Union[Unit, Application]) -> Diff:
     """
     # Retrieve the old data from the data key in the application relation databag.
     old_data = get_encoded_field(event.relation, bucket, "data")
+
+    if not old_data or (old_data and not isinstance(old_data, dict)):
+        # We should never get here, added to re-assure pyright
+        logger.error("Can't retrieve previous databag diff.")
+        old_data = {}
```
juditnovak, line 414: Aldready has a better shape locally (I just don't dare to ruin pipelines :sweat_smile: )

```
 411     if not old_data:
 412         old_data = {}
 413 
 414     if not isinstance(old_data, dict): 
 415         # We should never get here, added to re-assure pyright                                                                                                                                                 
 416         logger.error("Previous databag diff is of a wrong type.")
 417         old_data = {}
```

delgod (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -843,7 +993,10 @@ def update_relation_data(self, relation_id: int, fields: Dict[str, str]) -> None
         """Set values for fields not caring whether it's a secret or not."""
         relation = self.get_relation(self.relation_name, relation_id)
 
-        relation_secret_fields = get_encoded_field(relation, relation.app, REQ_SECRET_FIELDS)
+        if relation.app:
```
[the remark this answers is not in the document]
juditnovak: I'd (almost) say never, it's just linting that can't live without this extra security.
(But in reality I recall an execution that I came across where app was `None`. Can't remember if it was some artificial test case or not...?)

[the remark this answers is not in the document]
juditnovak: The primary reason why we have this extra check all across `data_interfaces.py` I still think is `pyright` though

welpaolo (approved):
Looks fine with me. 
I just left some minor comments related to type hints and docs. 

Did you consider to add some integration tests to validate those features ?

marceloneppel (approved):
Something important is the docstrings that Paolo mentioned.

juditnovak:
@welpaolo Regarding integration tests: 
```
kaboca@pegasus:~/repos/data-platform-libs$ grep -r fetch_relation tests/integration/ | grep -v data_interfaces | grep -v s3
kaboca@pegasus:~/repos/data-platform-libs$
```
Note that `fetch_relation_data()` has been there "forever"...

Personally I think that the unittests are actually pretty complete and conveniencing this time. Furthermore there are implicit tests as some of the internal worklfows of `data_intefaces.py` are actually using these functionalities. Typically by invoking the `is_resource_created()`, or `is_postgresql_plugin_enabled()` calls -- which are used in integration tests.

If we still prefer to have integration tests, I'm happy to add them, but for once I'd do it in a separate PR -- given that 3-4 important deliverables are now blocked missing the functionality provided by this PR here...

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -716,6 +808,50 @@ def fetch_relation_data(
...
+            .get(field)
+        )
+
+    def fetch_my_relation_data(
+        self,
+        relation_ids: Optional[List[int]] = None,
+        fields: Optional[List[str]] = None,
+        relation_name: Optional[str] = None,
+    ):
+        """Fetch data of the 'owner' (or 'this app') side of the relation."""
```
deusebio, line 827: Maybe it is worth adding in the docstring that this method should never be called from non-leader units. I feel that people may create code that uses this method without realizing that it should not be used outside of leaders. Maybe even re-naming `fetch_my_relation_data_from_leader` just to avoid this "common" mistake?

juditnovak: I may be missing something... Why would it be only the leader unit to call this function...?

We do not modify the relational data (which indeed is the privilege of the leader) but we're only reading it.

juditnovak: All right, summarizing offline discussion.

Apparently indeed, [by design](https://juju.is/docs/sdk/integration#heading--integration-data) only the leader unit has access to the `app` databag. Even READ access.

Thus this function can only return meaningful results if executed by the learder unit.

Adding the corresponding check.

juditnovak: Summarizing offline discussion:

While @deusebio has a valid point highlighting that *by design* [only the leader can access app data within a relation](https://juju.is/docs/sdk/integration#heading--integration-data), we still shouldn't restrain this function to the leader.

The reason is: every unit should be able to read their own relation data equally. Even if they don't have access to the `app` data.

This logic still has to be handled by the Charm Developer, outside of the generic interface equally provided here for both generic and leader units.

deusebio: Oh, wait, I may have misunderstood the scope, I thought that this method would be returning the application databag. if not, and if the function returns the unit databag nevermind, that can be resolved by both leaders and followers. 

So, just to confirm I understand correctly, can the method `fetch_my_relation_data` be called on both leaders and followers with no exceptions?

Either way, it is probably best to clarify what exactly the function returns in the docstrings, as it stands it is not clear whether it returns unit or app databags (or both?). Also, please add the typing for the output. I believe every public method should have that

delgod (approved):

welpaolo (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: All right, you are right, indeed we actually only return the Application databag (unit is ignored). In which case your observation is fully valid. Thanks for highlighting on it, confusion was on my side.

Added the change in a way that now there's also an ERROR in the logs once a leader-function was not invoked from a leader unit.

juditnovak:
For the record: since none of the leader-only functionaliites are confirmed with integration tests, we should address this issue.

I've opened Jira ticket https://warthogs.atlassian.net/browse/DPE-2730 (assigned to myself) to address this issue that goes back to the past.

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: That's great! Thanks! Could you also add the typing for the output and update the docstring (around the applcation data bag is returns and this should not be run as leader) and we'll be golden ;)

BTW...I really like the `@leader_only` decorator

juditnovak: Both done.

PR #100, opened 2023-10-13
[DPE-2746] Set relation alias also on application data level

dmitry-ratushnyy:
# Issue

When a relation alias is provided it should be set not only on the unit relation data level, but also on the application relation data level


delgod (approved):

marceloneppel (approved):

juditnovak (approved):
LGTM

PR #103, opened 2023-10-16
Check that a unit is leader before adding alias label on an application level

dmitry-ratushnyy:
# About

The goal of this PR is to fix error of changing application level relation data from non-leader units

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -320,7 +320,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 19
+LIBPATCH = 20
 
 PYDEPS = ["ops>=2.0.0"]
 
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1676,7 +1676,8 @@ def _assign_relation_alias(self, relation_id: int) -> None:
 
         # We need to set relation alias also on the application level so,
         # it will be accessible in show-unit juju command, executed for a consumer application unit
-        self.update_relation_data(relation_id, {"alias": available_aliases[0]})
+        if self.local_unit.is_leader():
+            self.update_relation_data(relation_id, {"alias": available_aliases[0]})
```
juditnovak, line 1680: Could you pls use the `leader_only` decorator instead?

juditnovak: (NOTE: Not on the event handler -- coz there we can't -- but on this function.

juditnovak (changes requested):

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Discussed offline: my advice above was silly (as I ignored the fact that the rest of the function is for all units :sweat_smile: )

However since `update_relation_data()` is restricted with the `leader_only` decorator, the `if` is not needed to be duplicated here.

delgod (approved):

marceloneppel (commented):
I see only the LIBPATCH bump in this PR. I think the second commit overwrote your changes.

marceloneppel (approved):

juditnovak (commented):
:+1:

juditnovak (approved):

PR #106, opened 2023-10-25
[DPE-2317] Small upgrade fixes

deusebio:
While testing upgrades on Kafka/Zookeeper VM, I found a couple of small fixes/improvements on UX. 

The most relevant one should be setting failed the unit if the upgrade_stack is not present. This used to create a sense of false expectation that upgrade run, even if the pre_upgrade action was not run and the upgrade_stack was not determined.

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -860,13 +860,9 @@ class DataUpgrade(Object, ABC):
 
         # if the first unit in the stack fails, the stack will be the same length as units
         # i.e this block not ran
-        if (
-            self.cluster_state in ["failed", "recovery"]
-            and self.upgrade_stack
-            and len(self.upgrade_stack) != len(self.app_units)
-            and self.charm.unit.is_leader()
-        ):
-            new_stack = self.upgrade_stack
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -1053,17 +1049,17 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
 
         if not self.upgrade_stack:
             logger.error("Cluster upgrade failed, ensure pre-upgrade checks are ran first.")
+            self.set_unit_failed("Ensure pre-upgrade checks are ran before upgrading.")
```
paulomach, line 1052: I don't think we should `set_unit_failed` here.
The idea being two fold: first, since there's no stopping a refresh, the charm should try to deal with it anyway; second, in k8s a pod reschedule/restart will trigger a `upgrade` event (ref. [juju bug #2021891](https://bugs.launchpad.net/juju/+bug/2021891) ) , so you can potentially set a `Blocked` status for the wrong reasons

paulomach (changes requested):

deusebio:
Closing as stable

PR #115, opened 2023-11-08
[DPE-2882] Demonstrating ValueError: Secret owner cannot use refresh=True

juditnovak:
Demonstrating use case for issue discussed in https://chat.charmhub.io/charmhub/pl/aekfocgk8tbtfcsda7311uurxe

THIS PR IS NOT TO BE MERGED


juditnovak:
POC, not needed anymore

PR #117, opened 2023-11-20
[DPE-3029] New common library -- secrets usage as shared across charms

juditnovak:
Extracting the common part of secrets usage, that's shared across the following charms (k8s + VM)

 - Mongo
 - Postgres
 - MySQL


delgod (approved):

lib/charms/data_platform_libs/v0/data_secrets.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Defined as a standalone function, as the choice on secret labels definition belongs to the
+    Application Logic. To be kept separate from classes below, which are simply to provide a
+    (smart) abstraction layer above Juju Secrets.
+    """
+    members = [charm.app.name, scope]
+    return f"{'.'.join(members)}"
+
+
+# Secret cache
```
marcoppenheimer, line 46: **nitpick:** This can be removed.

lib/charms/data_platform_libs/v0/data_secrets.py
```diff
@@ -0,0 +1,143 @@
...
+    The data structure is precisely re-using/simulating Juju Secrets behavior, while
+    also making sure not to fetch a secret multiple times within the same event scope.
+    """
+
+    def __init__(self, charm: CharmBase, label: str, secret_uri: Optional[str] = None):
+        self._secret_meta = None
+        self._secret_content = {}
+        self._secret_uri = secret_uri
+        self.label = label
+        self.charm = charm
```
marcoppenheimer, line 61: **question/thought:** Am I understanding correctly that if a charm-author re-initialises a cached secret, the `_secret_content` will be wiped, i.e not caching it? 
Although it's unlikely, is that behavior intended?

I wonder if using a singleton/[Borg](https://code.activestate.com/recipes/66531/) for the secrets is useful. Far less context on this though!

marcoppenheimer: Example Borg:

```python
class Borg:
    __shared_state = {}
    def __init__(self):
        self.__dict__ = self.__shared_state

    def add(self, arg):
        self.__shared_state[arg] = 123

a = Borg()
b = Borg()

a.add("1")
print(vars(b))
```
`>>> {"1": 123}`

lib/charms/data_platform_libs/v0/data_secrets.py
```diff
@@ -0,0 +1,143 @@
...
+        """Create a new secret."""
+        if self._secret_uri:
+            raise SecretAlreadyExistsError(
+                "Secret is already defined with uri %s", self._secret_uri
+            )
+
+        if scope == APP_SCOPE:
+            secret = self.charm.app.add_secret(content, label=self.label)
+        else:
+            secret = self.charm.unit.add_secret(content, label=self.label)
```
marcoppenheimer, line 73: **suggestion:** Not sure about how I feel about it, but we **could** do something like:

```python
secret = getattr(self.charm, scope).add_secret(content,label=self.label)
```

marcoppenheimer (approved):
Really clean and concise!
I left some unnecessary nits, and a (possibly uninformed) suggestion on how to cache across class instances, but nothing blocking.

lib/charms/data_platform_libs/v0/data_secrets.py
juditnovak: I know... I'm not sure which is worse :sweat_smile: 
Generally in most charms I've added helpers/properies to make similar decisions based on `Scope`.
Since in this lib it's a one-off, I'd just rather leave it as it is.

lib/charms/data_platform_libs/v0/data_secrets.py
juditnovak: I would leave it for now, just to separate --potentially future-- public functions, (vaguely) relating to the topic, from the actual Secret Caching itself...

PR #120, opened 2023-12-10
fix: add missing dependency (pylibjuju)

tonyandrewmeyer:
The tests use importlib.metadata to get the installed version of pylibjuju:

```python

However, pylibjuju is not installed as part of the requirements. This means that all tests fail with an error like:

```
ERROR tests/unit/test_data_interfaces.py::TestDatabaseProvides::test_database_requested_event - importlib.metadata.PackageNotFoundError: No package metadata was found for juju
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+juju ~= 3.1
 ops >= 2.1.1
 pydantic>=1.10,<2
 poetry-core

### CODE REVIEW

**Review by shayancanonical — APPROVED** [review]
TIL: tox env var substitutions. LGTM!

**Review by deusebio — APPROVED** [review]
Looks good! I would just update the reference to the juju to something more recent than 2.9.x

**Comment by tonyandrewmeyer** [comment]
> Looks good! I would just update the reference to the juju to something more recent than 2.9.x

It's the same as you have for your integration tests (I assumed you'd want it to match), but happy to change it to whatever you'd prefer.

**Comment by deusebio** [comment]
We tests a number of versions, among which 2.9.42.4 which is the oldest. I would just keep on testing on CI on that one, but I would move the default reference to latest, e.g. 3.5.0.0. Otherwise the reference would already seem out of date. 

I don't think that should break anything, transitive dependencies included in tox.ini are not resolved in charmlib anyway... :D

File: tox.ini [hunk 1]
```diff
@@ -71,6 +71,7 @@ deps =
     psycopg[binary]
     pytest<8.2.0
     pytest-mock
+    juju{env:LIBJUJU_VERSION_SPECIFIER:==2.9.42.4}
```
tonyandrewmeyer, line 74:
```suggestion
    juju{env:LIBJUJU_VERSION_SPECIFIER:==3.5.3}
```

PR #135, opened 2024-01-28
[DPE-3325] Dynamically adding Peer Data Secrets (groped on-demand)

juditnovak:
Most charms use a single secret on `app`, and another one on `unit` level. Therefore this was the functionality provided by the `DataPeer` and the `DataPeerUnit` interfaces.

This change is extending that behavior to what is only used by a few charms (Opensearch, Pgbouncer) allowing for multiple secrets on either level. The implemetation goes in a way, that new secret fields can also form new groups or be saved into an individual secret labelled after the key.


delgod (changes requested):
I need good case for absolutely dynamic secret allocation.

paulomach:
closing stale PRs

Issue #127, opened 2024-01-17
List of requested secrets MUST be possible to update dynamically -- for PeerData the least


juditnovak:
This missing functionality is blocking the usage of `DataPeer(Unit)` for `pgboucner`

PR #137, opened 2024-02-01
[DPE-3325] multi secrets dynamic update of fields only

juditnovak:
# Enhancements for Peer Relation Interfaces

As detailed in  [Peer Relation Data](https://docs.google.com/document/d/1folOFb--11xapLbrnmji7ErbLmagRffbYdiEPj_BTF8/edit?usp=sharing) spec.

## Changes

 - Re-introduce the concept of adding secrets on the fly.
   Up to this point `PeerRelation` and descendants only allowed for secret fields to be declared at instantiation time.
       - https://github.com/canonical/data-platform-libs/issues/127
 - Support for multiple secrets are supported STRICTLY if declared at instantiation time
    - Muiltiple secrets are currently only used by Opensearch that has multiple TLS secrets, however it's a valid scenario potentially for other charms as well in the future.
    - Since we want to strictly control the number of secrets "hanging around", dynamic secret creation is not allowed. A charm with multiple secrets should know precisely what additional secrets it may want to have further than the default ones.

## Terminology

Multiple secret groups may contain the same fields. Which means that we need to know which TLS `ca` do we want to get -- once we have 3 TLS secrets?  Therefore the following options were introduced:

- For the new `add_secret()` and `get_secret()` functions there is a `group` option to specify the group in case it's different from the default one

- For the unified `fetch_relation_data()`, `update_relation_data()`, etc. functions it may rather be desirable to keep the signature intact. Thus a `secret_field@group` (i.e. `ca-cert@tls-https`) notation is introduced. 

NOTE2: Implementation detail: this change still relies on the databag, however that's likely to change in the future, and secrets fields would be retrieved dynamically.


tests/integration/database-charm2/src/charm.py
```diff
@@ -0,0 +1,290 @@
...
+                    "startup": "enabled",
+                    "environment": {
+                        "PGDATA": "/var/lib/postgresql/data/pgdata",
+                        "POSTGRES_PASSWORD": self._stored.password,
+                    },
+                }
+            },
+        }
+        container.add_layer("database", pebble_layer, combine=True)
+        container.autostart()
```
marceloneppel, line 117: You don't need the workload here as you're not using it in the integration tests.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -337,21 +337,52 @@ def _on_topic_requested(self, event: TopicRequestedEvent):
...
 
-class SecretGroup(Enum):
-    """Secret groups as constants."""
+class SecretGroup(str):
+    """Secret groups specific type."""
 
-    USER = "user"
-    TLS = "tls"
-    EXTRA = "extra"
+    pass
```
marceloneppel, line 347: You can remove this as you have a docstring.
```suggestion
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -337,21 +337,52 @@ def _on_topic_requested(self, event: TopicRequestedEvent):
...
+            raise RuntimeError("Can't set constant!")
+        else:
+            super().__setattr__(name, SecretGroup(value))
+
+    def groups(self) -> list:
+        """Return the list of stored SecretGroups."""
+        return list(self.__dict__.values())
+
+    def get_group(self, group: str) -> Optional[SecretGroup]:
+        """If the input str translates to a grupname, return that."""
```
marceloneppel, line 370:
```suggestion
        """If the input str translates to a group name, return that."""
```

marceloneppel (commented):
I added some comments about details other than the lib update. I'll continue to review the lib changes tomorrow. Thanks for the PR, Judit.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -467,13 +502,46 @@ def wrapper(self, *args, **kwargs):
...
 
 
+def static_secrets_only(f):
+    """Decorator to ensure that certain operations would be only executed on Juju3."""
+
+    def wrapper(self, *args, **kwargs):
+        if self.static_secret_fields:
+            raise IllegalOperationError(
+                "Unsafe usage of statically and dynamically defined secrets, aborting."
+            )
```
marceloneppel, line 512: Shouldn't it be something like the code below?
```suggestion
        if set(self.current_secret_fields) - set(
            self.static_secret_fields
        ):
            raise IllegalOperationError(
                "Unsafe usage of dynamically defined secrets, aborting."
            )
```
It appears to me that this wrapper is currently raising an exception if we have static secrets.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -467,13 +502,46 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def static_secrets_only(f):
+    """Decorator to ensure that certain operations would be only executed on Juju3."""
```
marceloneppel, line 506: IMO, we need to change this docstring and the one from the next wrapper (`either_static_or_dynamic_secrets`) to match the check they are doing.

tests/integration/database-charm2/lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,3368 @@
+# Copyright 2023 Canonical Ltd.
```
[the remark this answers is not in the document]
juditnovak: This is only an initial copy. In fact, there is a section in `conftest.py`, that's taking care of copying the latest version of the lib into the charm.

The reason why the early presence of this file seems required are includes (an thus the static, initial copy). If I interpreted the failures (I received without this) correcty, includes may be evaluated before `conftests.py`. Therefore we may need the presence of this module before we could copy over the latest version.

If the above seem irrealistic, I can double-check if I may have made a mistake :-)

tests/integration/test_charm.py
```diff
@@ -390,6 +494,50 @@ async def test_peer_relation_non_leader_can_read_app_data(ops_test: OpsTest):
     assert action.results.get("value") == "plain text"
 
 
+@pytest.mark.abort_on_fail
+async def test_other_peer_relation(ops_test: OpsTest):
```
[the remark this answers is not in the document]
juditnovak: No, not on this one :-) 
This new interface is supposed to be fully functional regardless whether secrets are used or not.

And in fact it IS fully functional -- that's why both Juju 2 and Juju 3 pipelines are executing this test gree :-)

tests/integration/database-charm/src/charm.py
```diff
@@ -102,11 +104,40 @@ def __init__(self, *args):
...
+            return {}
+
+        for unit in self.peer_relation.units:
+            if unit not in self._servers_data:
+                self._servers_data[unit] = DataPeerOtherUnit(
+                    charm=self,
+                    unit=unit,
+                    relation_name=PEER,
+                    additional_secret_fields=UNIT_SECRETS,
+                )
```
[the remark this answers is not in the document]
juditnovak: I may be wrong here, but I see no issue with that...

It's kind of a general practice for attributes that are to be evaluated (constructed) only once during runtime, and not each time they would be referred to. Such as:
```
class X:
    def __init__(self):
        self._internal_heavy_calculation_value = {}
        
    @property
    def intrenal_heavy_calculation_value(self):
        if not self._internal_heavy_calculation_value:
            self._internal_heavy_calculation_value = some_heavy_calculation_done_here(self)
        return self._internal_heavy_calculation_value
```

In case I mis-interpreted your message, and this isn't what you meant: sry, it's my confusion.

tests/integration/database-charm2/src/charm.py
juditnovak: Yes, I know...
Originally I was thinking to have an alternative to the `database-charm` with the same functionality, for future extensions.
But I realize that all I foresee for the future is relations data related testing, which will never require a workload :sweat_smile:

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1645,23 +1735,167 @@ def scope(self) -> Optional[Scope]:
...
+        )
+
+    @property
+    def current_secret_fields(self) -> List[str]:
+        """Helper method to get all currently existing secret fields (added statically or dynamically)."""
+        if not self.secrets_enabled:
+            return []
+
+        if len(self._model.relations[self.relation_name]) > 1:
+            raise ValueError(f"More than one peer relation on {self.relation_name}")
```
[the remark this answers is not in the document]
juditnovak: Not according to Ian :-)  I asked this specifically

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Yes. And that's exactly the point.

See [DA076 - Peer Relation Data: No mixing of “dynamic” and “static” approach
](https://docs.google.com/document/d/1folOFb--11xapLbrnmji7ErbLmagRffbYdiEPj_BTF8/edit#heading=h.9e3keicxygz)

juditnovak: Bah, but I get your point -- the name of the decorator should be `dynamic_secrets_only` :sweat_smile: 
(NOT the opposite as it is currently: `static_secrets_only`)

Valid point :sweat_smile:

tests/integration/database-charm/src/charm.py
```diff
@@ -381,10 +413,11 @@ def _on_delete_peer_secret(self, event: ActionEvent):
             return
 
         secret = None
+        group_str = "" if not event.params["group"] else f".{event.params['group']}"
```
welpaolo, line 416: Is the dot in the f-string intentional?

tests/integration/dummy-database-charm/src/charm.py
```diff
@@ -0,0 +1,206 @@
...
+
+logger = logging.getLogger(__name__)
+
+PEER = "database-peers"
+
+
+class DatabaseCharm(CharmBase):
+    """Database charm that accepts connections from application charms."""
+
+    _stored = StoredState()
```
welpaolo, line 35: Do we really need this?

tests/integration/dummy-database-charm/src/charm.py
```diff
@@ -0,0 +1,206 @@
...
+            relation_name=PEER,
+            additional_secret_group_mapping={"mygroup": []},
+        )
+        self.peer_relation_unit = DataPeerUnit(self, relation_name=PEER)
+
+        # Default charm events.
+        self.framework.observe(self.on.start, self._on_start)
+
+        # Stored state is used to track the password of the database superuser.
+        self._stored.set_default(password=self._new_password())
```
welpaolo, line 52: And this?

tests/integration/dummy-database-charm/src/charm.py
```diff
@@ -0,0 +1,206 @@
...
+            self.on.change_admin_password_action, self._on_change_admin_password
+        )
+
+        self.framework.observe(self.on.set_secret_action, self._on_set_secret_action)
+
+        # Get/set/delete values on second-database relaton
+        self.framework.observe(
+            self.on.get_peer_relation_field_action, self._on_get_peer_relation_field
+        )
+        # self.framework.observe(
```
welpaolo, line 63: Do we want to keep this commented lines?

tests/integration/dummy-database-charm/src/charm.py
```diff
@@ -0,0 +1,206 @@
...
+                event.params["field"],
+                event.params["value"],
+                event.params["group"],
+            )
+
+    def _on_set_peer_secret(self, event: ActionEvent):
+        """Set requested relation field."""
+        component = event.params["component"]
+
+        # Charms should be compatible with old vesrions, to simulate rolling upgrade
```
welpaolo, line 152:
```suggestion
        # Charms should be compatible with old versions, to simulate rolling upgrade
```

tests/integration/dummy-database-charm/src/charm.py
```diff
@@ -0,0 +1,206 @@
...
+                event.params["value"],
+                event.params["group"],
+            )
+
+    # Remove peer secrets
+    def _on_delete_peer_secret(self, event: ActionEvent):
+        """Delete requested relation field."""
+        component = event.params["component"]
+
+        # Charms should be compatible with old vesrions, to simulate rolling upgrade
```
welpaolo, line 175:
```suggestion
        # Charms should be compatible with old versions, to simulate rolling upgrade
```

welpaolo (approved):
Looks good to me! 
I left some comments that are not blocking but would be nice to address

tests/integration/database-charm/src/charm.py
juditnovak: Yes, it is on purpose.

Secret labels are constructed such as `<application_name>.[app|unit][group_str], which maps to a `zookeeper.app.users` or `zookeeper.unit` structure

marceloneppel (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
marceloneppel: So, I believe this block can be removed to avoid any confusion in the future.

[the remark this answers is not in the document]
juditnovak: While I TOTALLY agree with you -- I'd still prefer to leave it as a security measure. 

I'm wondering: what if Juju changes something...? Or just may have a bug...? In that case our current workflow could become unreliable (insecure...?). I'd rather put a layer of security once we rely on a specific characteristic as such...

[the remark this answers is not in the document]
marceloneppel: It makes sense to me. Agreed.

PR #142, opened 2024-02-19
[DESIGN] Separation of Concerns (Data Abstraction vs. Event Handlers)

juditnovak:
Fully compatible with Charm Patters Design: https://github.com/canonical/zookeeper-operator/pull/121

(Backwards compatibility is confirmed by pipelines here.)

This PR implements the structure below. Allowing both for an internal separation of concerns (Data Abstraction vs. Event Handling), both for a smooth usage with [Data Interfaces Charm Patterns](https://docs.google.com/document/d/1TojdN0BN3qQ1fQ_uyMd7pCL6L5NNXtFo6DrOUwU5uFI/edit?usp=sharing) while keeping the old interface fully intact.

Data Interfaces version Charm Structure Patterns (3)

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -814,7 +814,7 @@ class Data(ABC):
 
         # If the relation started on a databag, we just stay on the databag
         # (Rolling upgrades may result in a relation starting on databag, getting secrets enabled on-the-fly)
-        # self.local_app is sufficient to check (ignored if Requires, never has secrets -- works if Provides)
+        # self.local_app is sufficient to check (ignored if Requires, never has secrets -- works if Provider)
         fallback_to_databag = (
             req_secret_fields
             and (self.local_unit == self._model.unit and self.local_unit.is_leader())
@@ -847,7 +847,7 @@ class Data(ABC):
 
         Since the Provider's databag is the only one holding secrest, we can apply
         a simplified workflow to read the Require's side's databag.
-        This is used typically when the Provides side wants to read the Requires side's data,
+        This is used typically when the Provider side wants to read the Requires side's data,
         or when the Requires side may want to read its own data.
         """
         if component not in relation.data or not relation.data[component]:
@@ -871,8 +871,8 @@ class Data(ABC):
 
         This function has internal logic to resolve if a requested field may be "hidden"
         within a Relation Secret, or directly available as a databag field. Typically
-        used to read the Provides side's databag (eigher by the Requires side, or by
-        Provides side itself).
+        used to read the Provider side's databag (eigher by the Requires side, or by
+        Provider side itself).
         """
         result = {}
         normal_fields = []
@@ -891,7 +891,7 @@ class Data(ABC):
             )
 
         # Processing "normal" fields. May include leftover from what we couldn't retrieve as a secret.
-        # (Typically when Juju3 Requires meets Juju2 Provides)
+        # (Typically when Juju3 Requires meets Juju2 Provider)
         if normal_fields:
             result.update(
                 self._fetch_relation_data_without_secrets(component, relation, list(normal_fields))
@@ -1072,10 +1072,10 @@ class EventHandlers(Object):
         raise NotImplementedError
 
 
-# Base ProvidesData and RequiresData
+# Base ProviderData and RequiresData
 
 
-class ProvidesData(Data):
+class ProviderData(Data):
     """Base provides-side of the data products relation."""
 
     def __init__(
@@ -1218,9 +1218,9 @@ class ProvidesData(Data):
     def _fetch_specific_relation_data(
         self, relation: Relation, fields: Optional[List[str]]
     ) -> Dict[str, str]:
-        """Fetching relation data for Provides.
+        """Fetching relation data for Provider.
 
-        NOTE: Since all secret fields are in the Provides side of the databag, we don't need to worry about that
+        NOTE: Since all secret fields are in the Provider side of the databag, we don't need to worry about that
         """
         if not relation.app:
             return {}
@@ -1309,8 +1309,8 @@ class ProvidesData(Data):
     fetch_my_relation_field = leader_only(Data.fetch_my_relation_field)
 
 
-class RequiresData(Data):
-    """Requires-side of the relation."""
+class RequirerData(Data):
+    """Requirer-side of the relation."""
 
     SECRET_FIELDS = ["username", "password", "tls", "tls-ca", "uris"]
 
@@ -1347,7 +1347,7 @@ class RequiresData(Data):
         then <arbitraty_label> will be "stuck" on the Secret object, whenever it may
         appear (i.e. as an event attribute, or fetched manually) on future occasions.
 
-        This will allow us to uniquely identify the secret on Provides side (typically on
+        This will allow us to uniquely identify the secret on Provider side (typically on
         'secret-changed' events), and map it to the corresponding relation.
         """
         label = self._generate_secret_label(relation_name, relation_id, group)
@@ -1429,7 +1429,7 @@ class RequiresData(Data):
     def _fetch_specific_relation_data(
         self, relation, fields: Optional[List[str]] = None
     ) -> Dict[str, str]:
-        """Fetching Requires data -- that may include secrets."""
+        """Fetching Requirer data -- that may include secrets."""
         if not relation.app:
             return {}
         return self._fetch_relation_data_with_secrets(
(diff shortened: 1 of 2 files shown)


welpaolo (approved):
The work here looks solid and consistent with the newly proposed charm pattern.

deusebio:
Overall, I really like the new structure with separation of concerns. I have some points worth raising and addressing to just nail this implementation properly. Given the large refactoring, I have been reviewing the entire file, not the diff. I find it easier to provide a general comment with references of the code commit therein, instead of inline comments of the diffs. Hence the long PR comment below. 

1. `DataProvides` and `DataRequires` still have some references to events [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1074) and [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1334). I don't see why having this method is necessary. This method feels more appropriate for `EventHandlers` classes, but actually I checked the code and we could probably just live without it, and replace the calls `self.relation_data._diff(event)` with `diff(event, self.relation_data.local_unit)` or `diff(event, self.relation_data.local_app)`, depending on the case. 

2. It does not seem to me that `DataRelation` classes need `charm` but only `model`, as it is clear from [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L631). I don't think we need to be backwards compatible on `DataRelation` (since this is actually a new abstraction of this PR), therefore I would just provide a constructor with `model`. Given these classes need to abstract away relation data, I suppose that model is good enough, rather than the entire charm. We anyway instantiate these classes in the final classes (e.g. [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L2462) where we could just put `charm.model` instead of `charm`). I would try to be a bit more restrictive and not pass always charm around, but only when really needed (e.g. handlers where we need to registrer observers).   

3. I would honestly have a general class that provides general functions for Provider and Requirer side, with no assumption on the databag payload/model underneath. As it is right now, `DataRequires` and `DataProvides` makes some assumption around some of the fields (e.g. [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L618), [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1272) and other methods right after, [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1312) and [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1318)  
3). This creates a somewhat dirty structure for some of the child classes that do not share the same data-model (e.g. Kafka, OpenSearch). Moreover, if we had such a class, we could use EXACTLY the same Provider and Requirer parents for S3, which right now does not support secrets and/or a bit more consistent UX, leveraging on the same functionalities. Eventually, even other teams could import from such Provider/Requirer classes if they were more general. 

4. Similarly to 3, I would also not make assumption on DataPeerData, [here](https://github.com/canonical/data-platform-libs/blob/7f6be35be78558416411b8449c4de1898d6ab396/lib/charms/data_platform_libs/v0/data_interfaces.py#L1525)

## Naming convention

This is largely personal, but given the these names are going to stick (and the discussion we just had about naming), I would like to propose a bit of simplification and structure on the naming of classes. Overall, I kinda find names like `DataPeerData` or `DataRequiresData` a bit confusing. Also `DataRequires` seems a bit inconsistent with its child `DataRequiresData`. It seems to me more natural to build on names as we go, in a structured way (notation below: `NewName (OldName)` ) :

Inheritance tree for Data classes: `Data (DataRelation)` > `ProvidesData (DataProvides)` and `RequiresData (DataRequires)` > `<Product>RequiresData (DatabaseRequiresData, KafkaRequiresData)` and `<Product>Data (PeerData)` (when we put Requires and Provides together).   

Inheritance tree for EventHandlers classes: `EventHandlers (DataRelationEventHandlers)` > `ProvidesEventHandlers (DataProvidesEventHandlers)` and `RequiresEventHandlers (DataRequiresEventHandlers)` > `<Product>RequiresEventHandlers (DatabaseRequiresEventHandlers, KafkaRequiresEventHandlers, ...)`

I would have a niptick that we should probably use `Requirer` in place of `Requires` which would seem more appropriate from an English standpoint (e.g. `RequirerData` instead of `RequiresData`). We could probably do this while still retaining the final class naming, e.g. `class DatabaseRequires(DatabaseRequirerData, DatabaseRequirerEventHandlers)`, but I can see that this may hurt someone else. Anyhow, it is not a bit point. I would prefer to go with better English, but I don't mind too much.

juditnovak:
@deusebio Responses one-by-one prioritized by simplicity

(2) `Data` classes are only taking `Model` now. Very good call, this had to be cleaned up.

(1) `_diff()` is a tricky one -- as it is kind of a link between data and the way data is being used. It takes BOTH the event, AND the "databag-in-question" for the `Data` class in question.
Instead of the independent function, we _could_ move it to `DataRelation` -- where the corresponding `self.databag_in_question` could be determined or re-defined in descendants.
However, in this case it would be the `DataPeer` classes inheriting the whole stuff pointlessly.

To me the fact that a `Data` class has a method with an `Event` parameter in it is not necessarily a problem. After all these `Data` classes exist to be used in a context with events.

Alternatively we could re-define the signature of the `diff()` function to rather take a `DataRelation` instance as input instead of `Union[App, Unit]`. That way the event handler could call `diff(event, self.data). And I'm adding a new field on `DataRelation` such as `self._data_component` (as sadly `self.component` is NOT the same). 
How about this solution?

(3) This is an idea that worth consideration for sure. However I would evaluate and address it in a separate PR. First of all, Opensearch IS sharing the model (even if (currently?) Kafka may not). While your point is to have fully independent parent classes, perhaps for more generic future usage, as for now parents are also to define sensible default (that could be overriden in heirs, like `SECRET_FIELDS` if needed). To me this is an architectual question, that may worth considering. Yet I would do that separately, thus iteratively CI/CD-style improving -- instead of putting ALL our ideas into this PR.

(4) It is the default that a number of charms seem to have defined. (If not it's harmless to have). The charms ARE defining their own secrets using the `additional_secrets_fields` (for now -- that one may get a better name in the future). Removing the `operator-password` default may make sense, I'm thinking of it for a while -- however I'd keep that change independent from internal architectural changs. ALSO that's a STRONG behavioral change that has to be well communicated in advance.

***Naming***: The main rationale of the naming was to avoid overlap with Juju's `RelationData`. Thus I swapped the name to `DataRelation` -- and followed that structural logic consistently.
I'm not against renaming the internal classes such as 
 - `Data` (`DataRelation`)
 - `Provides` (`DataProvides`)
 - `ProvidesData` (`DataProvidesData`)
 - `<Product>`ProvidesData` -- remains
 - EventHandlers (DataRelationEventHandlers)
 - ProvidesEventHandlers (DataProvidesEventHandlers)
 
 I am against renaming ANY of the interface classses (such as `DataPeer`), as that's a breaking change for multiple charms.

juditnovak:
@deusebio 

I would prefer (3) to be discussed before implementing any code.

(4) I'm adding on this PR

juditnovak:
:exclamation: :exclamation: :exclamation: :exclamation: NOT CHANGING `_diff()` :exclamation: :exclamation: :exclamation: :exclamation: :exclamation: 

`_diff()` turns out to be used all across our charms. Whether that's good or bad, we should NOT breaking that (right here, right now the least)

juditnovak:
@deusebio @welpaolo @delgod 

Renaming and `Model`-related changes added in separate PRs.
Please let me know if you are happy with them or not. Separate PRs thus easy to remove if objections.

deusebio:
> There may be one risk of removing _diff() though.

Just to be safe and backward compatible (although we should put some deprecation message on `_diff` if it is not needed, and people in general should not use private methods actually...), one could also 

1. Keep the same signature for `diff` 
2. Create a `_diff` method in `EventHandlers` with this:

```
class EventHandlers:
    def _diff(self, event):
        return diff(event, self.relation_data.data_component)
```

Wouldn't this respect the same API for both global `diff` and `EventHandlers._diff`?

deusebio:
> Let me just confirm with Mykola and Paolo whether to use Requirer or Requires

I checked and other charming teams use Requirer/Provider terminology (see e.g. see [here](https://charmhub.io/oathkeeper/libraries/auth_proxy#authproxyprovider) and [here](https://charmhub.io/traefik-k8s/libraries/ingress#ingressperappprovider) coming from Observability and Identity), so we should also follow the same standard

welpaolo:
Hi @juditnovak, I am fine with the plan you proposed.

I will keep an eye on the removal of `diff()` function to ensure that all the tests works :)  

@deusebio In terms of naming let's follow what others teams are doing and be compliant with the rest of the repositories.

delgod:
* Feel free to remove `diff()` if our charms don't use it.
* About naming, let's follow ecosystem naming.

deusebio (approved):
Looks very good! Thanks!

welpaolo (approved):

PR #143, opened 2024-02-21
[POC] "Back to databag" (i.e. `dict` representation of Abstract Relation Data

juditnovak:
See how it works on ZooKeeper: https://github.com/canonical/zookeeper-operator/pull/122

If the concept is welcome, I'm adding the tests

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -558,6 +558,18 @@ class CachedSecret:
         if self.meta:
             return self.meta.get_info()
 
+    def remove(self) -> None:
+        """Remove secret."""
+        if not self.meta:
+            raise SecretsUnavailableError("Non-existent secret was attempted to be removed.")
+        try:
+            self.meta.remove_all_revisions()
+        except SecretNotFoundError:
+            pass
+        self._secret_content = {}
+        self._secret_meta = None
+        self._secret_uri = None
+
 
 class SecretCache:
     """A data structure storing CachedSecret objects."""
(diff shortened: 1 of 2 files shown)


juditnovak:
Summary of the offline conversation with @welpaolo : 

 - We evaluated what if we use direct inheritace from `dict` for `*Data` classes. i.e. no `*Data.as_dict()` method (as the POC has it at the time this comment is written).
   We agreed that it seems a good idea, as this allows for instantiating the object directly, instead of requiring a second object to benefit from `dict` access (i.e. `X` instance, `Y = X.as_dict()`).
 - However, the above means a bigger, conceptual decision.
   I.e. that Python Standard Library `dict` would become our confirmed interface for Abstract Relation Data.
   - @welpaolo had once concern: In case Python Standard Lib introduces a (breaking) change, we must follow -- and/or we might need to pin Python versions over the lib
      - Though the above is a valid concern, in my opinion if `dict` has a (breaking) change, it's likely to have a strong impact on the rest of our code anyway. So adjusting Relation Data may just be part of handling the problem overall.
      - Since no matter what `dict` IS a key-value mapping, it should ever correspond to the representation of Relation Data from our perspective. Thus overall it should be convenient to be used as an interface.

juditnovak:
@welpaolo 
I was an idiot. (Let's hope it's past tense :rofl: )

We need two separate objects. 

`Data` (here still called `DataRelation`) is bound to a relation-name.

While the `dict` is representing a specific relation within that "name".

So we can't do better than `as_dict(relation_id)` :-/ 

Sorry I was thinking on a wrong track :-/

PR #147, opened 2024-03-21
[DPE-3874] `dict` representation for Data (i.e. "back to databag" :-) )

juditnovak:
This PR is the "continuation" of [earlier POC](https://github.com/canonical/data-platform-libs/pull/143).

See how it works on ZooKeeper: https://github.com/canonical/zookeeper-operator/pull/122

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -295,10 +295,21 @@ import copy
 import json
 import logging
 from abc import ABC, abstractmethod
-from collections import namedtuple
+from collections import UserDict, namedtuple
 from datetime import datetime
 from enum import Enum
-from typing import Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import (
+    Callable,
+    Dict,
+    ItemsView,
+    KeysView,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+    ValuesView,
+)
 
 from ops import JujuVersion, Model, Secret, SecretInfo, SecretNotFoundError
 from ops.charm import (
@@ -320,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 30
+LIBPATCH = 31
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -610,6 +621,102 @@ class SecretCache:
 
 
 # Base Data
+
+
+class DataDict(UserDict[str, str]):
+    """Python Standard Library 'dict' - like representation of Relation Data."""
+
+    def __init__(self, relation_data: "Data", relation_id: int):
+        self.relation_data = relation_data
+        self.relation_id = relation_id
+
+    @property
+    def data(self) -> Dict[str, str]:
+        """Return the full content of the Abstract Relation Data dictionary."""
+        result = self.relation_data.fetch_my_relation_data([self.relation_id])
+        try:
+            result_remote = self.relation_data.fetch_relation_data([self.relation_id])
+        except NotImplementedError:
+            result_remote = {self.relation_id: {}}
+        if result:
+            result_remote[self.relation_id].update(result[self.relation_id])
+        return result_remote.get(self.relation_id, {})
+
+    def __setitem__(self, key: str, item: str) -> None:
+        """Set an item of the Abstract Relation Data dictionary."""
+        self.relation_data.update_relation_data(self.relation_id, {key: item})
+
+    def __getitem__(self, key: str) -> str:
+        """Get an item of the Abstract Relation Data dictionary."""
+        result = None
+        if not (result := self.relation_data.fetch_my_relation_field(self.relation_id, key)):
+            try:
+                result = self.relation_data.fetch_relation_field(self.relation_id, key)
+            except NotImplementedError:
+                pass
+        if not result:
+            raise KeyError
+        return result
+
+    def __eq__(self, d: dict) -> bool:
+        """Equality."""
+        return self.data == d
+
+    def __repr__(self) -> str:
+        """String representation Abstract Relation Data dictionary."""
+        return repr(self.data)
+
+    def __len__(self) -> int:
+        """Length of the Abstract Relation Data dictionary."""
+        return len(self.data)
+
+    def __delitem__(self, key: str) -> None:
+        """Delete an item of the Abstract Relation Data dictionary."""
+        self.relation_data.delete_relation_data(self.relation_id, [key])
+
+    def has_key(self, key: str) -> bool:
+        """Does the key exist in the Abstract Relation Data dictionary?"""
+        return key in self.data
+
+    def update(self, items: Dict[str, str]):
+        """Update the Abstract Relation Data dictionary."""
+        self.relation_data.update_relation_data(self.relation_id, items)
+
+    def keys(self) -> KeysView[str]:
+        """Keys of the Abstract Relation Data dictionary."""
+        return self.data.keys()
+
+    def values(self) -> ValuesView[str]:
+        """Values of the Abstract Relation Data dictionary."""
+        return self.data.values()
+
+    def items(self) -> ItemsView[str, str]:
+        """Items of the Abstract Relation Data dictionary."""
+        return self.data.items()
+
+    def pop(self, item: str) -> str:
+        """Pop an item of the Abstract Relation Data dictionary."""
+        result = self.relation_data.fetch_my_relation_field(self.relation_id, item)
+        if not result:
+            raise KeyError(f"Item {item} doesn't exist.")
+        self.relation_data.delete_relation_data(self.relation_id, [item])
+        return result
+
+    def __contains__(self, item: str) -> bool:
+        """Does the Abstract Relation Data dictionary contain item?"""
+        return item in self.data.values()
+
+    def __iter__(self):
+        """Iterate through the Abstract Relation Data dictionary."""
+        return iter(self.data)
+
+    def get(self, key: str, default: Optional[str] = None) -> Optional[str]:
+        """Safely get an item of the Abstract Relation Data dictionary."""
+        try:
+            if result := self[key]:
+                return result
+        except KeyError:
+            return default
 
 
 class Data(ABC):
@@ -928,6 +1035,10 @@ class Data(ABC):
 
     # Public interface methods
     # Handling Relation Fields seamlessly, regardless if in databag or a Juju Secret
+
+    def as_dict(self, relation_id: int) -> UserDict[str, str]:
+        """Dict behavior representation of the Abstract Data."""
+        return DataDict(self, relation_id)
 
     def get_relation(self, relation_name, relation_id) -> Relation:
         """Safe way of retrieving a relation."""
--- a/tests/unit/test_data_interfaces.py
+++ b/tests/unit/test_data_interfaces.py
@@ -107,6 +107,141 @@ def verify_relation_interface_functions(interface, relation_id):
     assert rel_data[relation_id] == {}
 
 
+def verify_relation_interface_dict(interface, relation_id):
+    interface_dict = interface.as_dict(relation_id)
+    orig_len = len(interface_dict)
+    orig_dict = dict(interface_dict)
+
+    # Interface function: update_relation_data()
+    interface_dict["something"] = "else"
+
+    # Interface function: fetch_relation_field()
+    assert interface_dict["something"] == "else"
+    assert len(interface_dict) == orig_len + 1
+
+    # Interface function: pop()
+    assert interface_dict.pop("something") == "else"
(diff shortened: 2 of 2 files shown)


welpaolo (commented):
The implementation looks good. I feel there is the need for some integration tests tests to check the handling of relation data as dictionary. 

Do you think it will be possible (just for testing purposes) to have "normal" and dict representation to work in the same testing charm?

juditnovak:
@welpaolo your question/request about integration tests is indeed something worth discussing. 
I apologize, my response is long, as to me it's a _conceptual_ matter (not in terms of code but in terms of tests). 
I'm glad to sync on this after your holidays if preferred -- as we usually do :-) 

In fact, I was thinking to add a comment on this topic to the **Description**, when I created the PR, but I didn't do it at the end.

There are  purposefully no integration tests here. I don't think we should have them this time.
However, there may be something else to be tested instead -- that is missing currently. I rather propose that extension (see later).

The `dict` interface should be VERY strictly no more than a "syntactic sugar", a lightweight wrapper around the 3(+1) interface functions we are providing:
 - `fetch_relation_data()` (and `fetch_relation_field()`
 - `update_relation_data()`
 - `delete_relation_data()`

These 3 are implementing the _real_ functionality and must be particularly carefully tested with integration tests. (I think we're pretty good on that right now.)

The `dict` wrapper MUSTN'T invent anything. It should add NO new functionality. It must _strictly_ do no more than apply the underlying interface function that corresponds to the `dict` call, with the parameters that were shipped.

To test this lightweight wrapper layer (new interface), I see no value in integration tests. ALL that the wrapper has to provide is rather a unit test material.

However I do see a missing element in the current unit tests indeed. They do verify every aspect of the `dict` behavior, however they _don't_ ensure that it is no more than a thin wrapper around the interface functions.
(While it is VERY important to ensure that further additions to the code would preserver the wrapper nature, i.e. "no functionality would be invented" within a `DataDict` object.)

Thus, what I rather propose is that I would extend the current unittests with one more, that is verifying that the `dict` wrapper is calling the underlying interface functions. (IT is one of the few cases where I find mocks being the right thing to use.)
Also I would add an explanation (both on the code (`class`) an in the tests) how this concept MUST be preserved.

[NOTE: While I'm VERY keen on through testing, I also think it's important to write tests thoughtfully, in a sense to make sure that we are testing the right thing in the right place, at the right time. Especially when dealing with complex code, tests can get confusing, difficult to extend, difficult to "debug", loose descriptive nature if not developed carefully.]

One more reason why I find it more beneficial not to add another test charm, is because we _will_ have a new test charm in [my upcoming, heavyweight change](https://github.com/canonical/data-platform-libs/pull/137) -- which _will_ require a new test charm :-) 
(Sadly it won't be convenient to provide a full test coverage for the `dict` usage there, but if preferred I can use the `dict` syntax in that one :-) So that `data-platform-libs` provide  a charm using the syntax.). 
But I think it may keep the testsuite more clean if we don't add too many test charms... as it may get confusing, difficult to follow...

(BTW regarding "real-life" functionality: of course this code is being used in the PR mentioned in the **Description**, functioning real well :-) Just to reassure that the proposal has been executed by actual charms, showing good results :-) )

Huh, I apologize for writing a novel here... Testing is a very important topic for me :sweat_smile:

juditnovak:
Hmmmmm.... And another missing thing here :sweat_smile: 

Those unittests we should have on the Cross-charm relation interfaces as well :facepalm: 

I'm adding this missing bit as well!!!!!

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -929,6 +1024,10 @@ def _delete_relation_data_without_secrets(
     # Public interface methods
     # Handling Relation Fields seamlessly, regardless if in databag or a Juju Secret
 
+    def as_dict(self, relation_id: int):
```
deusebio, line 1027: **todo** I blieve it is safe to type it as

```
def as_dict(self, ...) -> dict:
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
 # Base Data
 
 
+class DataDict(dict):
```
deusebio, line 615: Uhm, this only extends `Dict[str, str]` or any dict?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+        """Does the key exist in the Abstract Relation Data dictionary?"""
+        return k in self.all_data()
+
+    def update(self, items: Dict[str, str]):
+        """Update the Abstract Relation Data dictionary."""
+        self.relation_data.update_relation_data(self.relation_id, items)
+
+    def keys(self) -> List[str]:
+        """Keys of the Abstract Relation Data dictionary."""
+        return list(self.all_data().keys())
```
deusebio, line 675: To just keep closer to original implementation, we can probably avoid the list conversion, and just type with common keys() output

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
 
 
+class DataDict(dict):
+    """Python Standard Library 'dict' - like representation of Relation Data."""
+
+    def __init__(self, relation_data: "Data", relation_id: int):
+        self.relation_data = relation_data
+        self.relation_id = relation_id
+
+    def all_data(self) -> Dict[str, str]:
```
deusebio, line 622:
```suggestion
    @property
    def data(self) -> Dict[str, str]:
```
```suggestion
    def all_data(self) -> Dict[str, str]:
```

**niptick**

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+
+    def update(self, items: Dict[str, str]):
+        """Update the Abstract Relation Data dictionary."""
+        self.relation_data.update_relation_data(self.relation_id, items)
+
+    def keys(self) -> List[str]:
+        """Keys of the Abstract Relation Data dictionary."""
+        return list(self.all_data().keys())
+
+    def values(self) -> List[str]:
```
deusebio, line 677: as above

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+        """Values of the Abstract Relation Data dictionary."""
+        return list(self.all_data().values())
+
+    def items(self) -> List[Tuple[str, str]]:
```
deusebio, line 681: as above

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+
+    def __len__(self) -> int:
+        """Length of the Abstract Relation Data dictionary."""
+        return len(self.all_data())
+
+    def __delitem__(self, key: str) -> None:
+        """Delete an item of the Abstract Relation Data dictionary."""
+        self.relation_data.delete_relation_data(self.relation_id, [key])
+
+    def has_key(self, k) -> bool:
```
deusebio, line 665:
```suggestion
    def has_key(self, key: str) -> bool:
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+                result = self.relation_data.fetch_relation_field(self.relation_id, key)
+            except NotImplementedError:
+                pass
+        if not result:
+            raise KeyError
+        return result
+
+    def __eq__(self, d: dict) -> bool:
+        """Equality."""
+        return set(self.keys()) == set(d.keys()) and all(self[key] == d[key] for key in d)
```
deusebio, line 651: **question** why not using built-in equality for dicts `self.data == d`? Note sure I was able to understand why elsewhere you used built-in implementations (e.g. `__repr__`) and not here

deusebio (changes requested):
Given this is an implementation to keep as close as possible to `dict`, I've been a bit more picky (and we should probably be) for the typing and naming used in the signatures.

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: I meant to implement the `dict[str,str]`, given we know that keys and values can only be strings, i.e.

```
class DataDict(dict[str, str]):
```

By the way, it is also probably better to extend the `UserDict` class rather than `dict` (see [here](https://stackoverflow.com/questions/7148419/subclass-dict-userdict-dict-or-abc)), so 

```
class DataDict(UserDict[str, str]):
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -612,6 +612,101 @@ def remove(self, label: str) -> None:
...
+            result_remote = {self.relation_id: {}}
+        if result:
+            result_remote[self.relation_id].update(result[self.relation_id])
+        return result_remote.get(self.relation_id, {})
+
+    def __setitem__(self, key: str, item: str) -> None:
+        """Set an item of the Abstract Relation Data dictionary."""
+        self.relation_data.update_relation_data(self.relation_id, {key: item})
+
+    def __getitem__(self, key: str) -> Optional[str]:
```
marcoppenheimer, line 637: **todo:** I don't think this can be 'Optional'. We should raise a KeyError if we can't find the key, like a normal dict.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: All right, I see your point. Sorry, it's a very valid consideration. My bad, I wasn't aware of generics here. Thanks for pointing it out, it was very good info.

I'm thinking of this, and... I'm not sure what may be the best here.

On one hand, I'm thinking if we may want to keep the implementation flexible (`Dict[str, Any]`?), typically in case in the future, for any unforeseen reason we may want to have more complex structures (`list`, `dict`, etc.) as values. It may come handy not to restrict more than necessary.

(Of course, on the other hand it's true that current `get()`, `__getitem__()`, etc. are returning `str` only... We may want to keep that on `Any` as well...?)

Thinking.... Pro-s/con-s are welcome.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: It actually does, my mistake. Changed accordingly.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Changed to `UserDict`, kept in-line with the discussion below.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
juditnovak: Done.

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: I believe it should be `[str, str]`, as that is the only type allowed for data bags. This data structure is pretty low-level and close to the databag, so I would keep the signature aligned. If someone wants to create more "structure", like simple types (int, float, list, ...) but even more complex with user-derived dataclasses, they should probably implement serialization/deserialization starting from the functionalities exposed by this class (that are with strings)

deusebio (approved):
LGMT! Thanks!

welpaolo (approved):
LGTM

PR #150, opened 2024-03-21
[DPE-3876][BUGFIX] DataPeerOtherUnit

juditnovak:
There was a major copy-paste mistake on the new `DataPeerOtherUnit` class (that will be required typically for [this functionality](https://github.com/canonical/zookeeper-operator/blob/c71086ffe0cfb877c5aa11b99765cc6e87d26db9/src/core/cluster.py#L59-L70) demonstarted in Zookeeper).

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -320,7 +320,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 29
+LIBPATCH = 30
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -1787,6 +1787,14 @@ class DataPeerOtherUnitData(DataPeerUnitData):
         self.local_unit = unit
         self.component = unit
 
+    def update_relation_data(self, relation_id: int, data: dict) -> None:
+        """This method makes no sense for a Other Peer Relation."""
+        raise NotImplementedError("It's not possible to update data of another unit.")
+
+    def delete_relation_data(self, relation_id: int, fields: List[str]) -> None:
+        """This method makes no sense for a Other Peer Relation."""
+        raise NotImplementedError("It's not possible to delete data of another unit.")
+
 
 class DataPeerOtherUnitEventHandlers(DataPeerEventHandlers):
     """Requires-side of the relation."""
@@ -1809,10 +1817,10 @@ class DataPeerOtherUnit(DataPeerOtherUnitData, DataPeerOtherUnitEventHandlers):
         additional_secret_fields: Optional[List[str]] = [],
         secret_field_name: Optional[str] = None,
         deleted_label: Optional[str] = None,
-        unique_key: str = "",
     ):
-        DataPeerData.__init__(
+        DataPeerOtherUnitData.__init__(
             self,
+            unit,
             charm.model,
             relation_name,
             extra_user_roles,
@@ -1820,7 +1828,7 @@ class DataPeerOtherUnit(DataPeerOtherUnitData, DataPeerOtherUnitEventHandlers):
             secret_field_name,
             deleted_label,
         )
-        DataPeerEventHandlers.__init__(self, charm, self, unique_key)
+        DataPeerOtherUnitEventHandlers.__init__(self, charm, self)
 
 
 # General events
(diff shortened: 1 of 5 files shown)


tests/integration/test_charm.py
```diff
@@ -390,6 +393,48 @@ async def test_peer_relation_non_leader_can_read_app_data(ops_test: OpsTest):
     assert action.results.get("value") == "plain text"
 
 
+@pytest.mark.abort_on_fail
+async def test_other_peer_relation(ops_test: OpsTest):
```
welpaolo, line 397: As an additional tests it will really nice to have the following:
- Scale down the charm to 1 or 2 units and check that the configured items are still there
- Scale up and check that the additional unit does not have any item in the databag

tests/unit/test_data_interfaces.py
```diff
@@ -80,19 +82,67 @@
 """
 
 
+#
+# Helper functions
+#
+
+
+def verify_relation_interface_functions(interface, relation_id):
```
welpaolo, line 90: Please add docstring

tests/unit/test_data_interfaces.py
```diff
@@ -80,19 +82,67 @@
...
+    # update_relation_data()
```
welpaolo, line 91: Are these following commented lines left here on purpose?

welpaolo (commented):
Looks a good fix. I left some comments

tests/unit/test_data_interfaces.py
juditnovak: Yes, I was meant to indicate which interface function is being tested at which part of the test.

I'll change the comment to make this obvious, indeed now it seems like a commented action :sweat_smile:

deusebio: Yes, probably with the docstring would address this one then

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1787,6 +1787,14 @@ def __init__(self, unit: Unit, *args, **kwargs):
         self.local_unit = unit
         self.component = unit
 
+    def update_relation_data(self, relation_id: int, data: dict) -> None:
+        """This method makes no sense for a Other Peer Relation."""
+        raise NotImplementedError("It's not possible to update data of another unit.")
+
+    def delete_relation_data(self, relation_id: int, fields: List[str]) -> None:
+        """This method makes no sense for a Other Peer Relation."""
+        raise NotImplementedError("It's not possible to delete data of another unit.")
```
deusebio, line 1796: **thought** I believe I have said this already but possibly the use-case of having these methods could rather call for an abstraction between writable and non-writable data in the Data object. I generally recommend to rather not have methods over raising `NotImplementedError` (for concrete methods, as above), as the second would only appear at runtime in (possibly) some conditions/corner cases, whereas the former will not even allow the developer to write the code (since mypy would complain), and therefore it should be safer.

deusebio (approved):
The changes look fine, although I'm inclined to think that this calls for thinking a bit the abstraction that we provide, to provide more typesafe code

tests/integration/test_charm.py
```diff
@@ -57,7 +57,7 @@ async def test_deploy_charms(ops_test: OpsTest, application_charm, database_char
                 ]
             },
             application_name=DATABASE_APP_NAME,
-            num_units=2,
+            num_units=3,
```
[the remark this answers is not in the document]
juditnovak: Normally we are running tests on applications with 3 nodes. Somewhat 2 sounds a bit too "few", still close to a corner-case. 

I actually think that we used to have 3 nodes here that I change for my local tests to 2 each time (min. required for :green_circle: )

I think I accidentally decreased this number on a previous PR by committing my local change :sweat_smile:

tests/integration/test_charm.py
juditnovak: Done.

tests/unit/test_data_interfaces.py
juditnovak: Done.

tests/unit/test_data_interfaces.py
juditnovak: Done.

juditnovak:
@welpaolo @deusebio +1 commit added (https://github.com/canonical/data-platform-libs/pull/150/commits/1758e05db76de9ef8253aa07d156e2fdeac29027) with extending interfaces unittests to all interfaces.

This is slightly redundant, as most (BUT not all!!!) of this functionality has been tested specifically for various use cases. Yet this "consistent redundancy" won't hurt. (Rather the opposite: it will cover missing bits.)

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: Yes, it is DEFINTELY out of the scope.

welpaolo (approved):
Looks good to me!

Issue #152, opened 2024-04-03
[BUGFIX] Issue with `dict` representation


juditnovak:
## Steps to reproduce

Subscripting not supported, see https://github.com/canonical/indico-operator/actions/runs/8534721518/job/23380023282?pr=374#step:20:59

See all details on the pipeline above (versions, logs, etc.).

## Additional context

Matrix discussion: https://matrix.to/#/!BukWfnyOTgQSKAxdtT:ubuntu.com/$X0SV2M_Xb-MYL3gOAmXJzuFVSOeFQc9MKXBIwDUwA6w?via=ubuntu.com&via=matrix.org

PR #153, opened 2024-04-04
[DPE-3906][BUGFIX] dict implementation not supporting Python 3.9> (issue #152)

juditnovak:
# URGENT FIX -- Python 3.9> compatiblity broken

Once again I had to remember the hard way: `data_interfaces` MUST be strongly backwards compatible.

Both in terms of code and in terms of Python versions, etc.

Adding generics to the `dict` implementation parent class broke Python 3.9> compatibility.

The proposed fix executes correctly on earlier Python versions as well: https://github.com/canonical/data-platform-libs/actions/runs/8559046518

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 32
+LIBPATCH = 33
 
 PYDEPS = ["ops>=2.0.0"]
 
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -690,7 +690,7 @@ class SecretCache:
 # Base Data
 
 
-class DataDict(UserDict[str, str]):
+class DataDict(UserDict):
     """Python Standard Library 'dict' - like representation of Relation Data."""
 
     def __init__(self, relation_data: "Data", relation_id: int):
@@ -1105,7 +1105,7 @@ class Data(ABC):
     # Public interface methods
     # Handling Relation Fields seamlessly, regardless if in databag or a Juju Secret
 
-    def as_dict(self, relation_id: int) -> UserDict[str, str]:
+    def as_dict(self, relation_id: int) -> UserDict:
         """Dict behavior representation of the Abstract Data."""
         return DataDict(self, relation_id)
 


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -623,7 +623,7 @@ def remove(self, label: str) -> None:
 # Base Data
 
 
-class DataDict(UserDict[str, str]):
```
juditnovak, line 626: Note: I know there are ways to have the generics selective to version.

However I'm wondering to rather not enforce typing here, leaving it open for potential future extensions.

marceloneppel: Indico build still fails due to generics at `def as_dict(self, relation_id: int) -> UserDict[str, str]:` (line 1039).

marceloneppel (changes requested):

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: GREAT CATCH THANK U SO MUCH!!!!!!!!!!!

deusebio (approved):
That's a pity we are stuck to be compatible to old version of python. Why are people still using 3.8?!? EOL is coming up this year. 

**niptick** I feel the title and the commit reads better to say "Python <3.9"

marcoppenheimer:
I have responded specifically, but I don't think we should support non-latest-LTS Ubuntu Python versions.

juditnovak:
@marceloneppel , @deusebio  I don't think we should limit our lib users in their choice of Ubuntu/Python version. We have many charms (like wordpress or indico). Neither we should deprive latest enhancements (or bugfixes) from them. 

We can have pretty, up-to-date code in our charms where we're in control of Ubuntu/Python versions, where we know what our clients need/use, etc. But generic interfaces that we provide should be more allowing than that.

As explained in the [related issue](https://github.com/canonical/data-platform-libs/issues/152), the problem was raised on the public Matrix channel by the Indico charm team: https://matrix.to/#/!BukWfnyOTgQSKAxdtT:ubuntu.com/$X0SV2M_Xb-MYL3gOAmXJzuFVSOeFQc9MKXBIwDUwA6w?via=ubuntu.com&via=matrix.org

deusebio:
To me, EoL is a good enough reason to stop making sure code is backward compatible with that version. Otherwise how do we draw the line between which versions you want to be compatible with?

Anyhow, I would probably add a matrix on the CI (only unittests in my opinion) against the different python versions we support. Of course, outside of this PR if the other team is blocked and we need to merge this asap.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done

juditnovak:
Let's be honest -- we introduced a breaking change ;-) 
While indeed we may want to consider what level of backwards compatibility should our libs support -- equally we should avoid incompatible/breaking changes :-)  At least across minor versions

juditnovak:
Adding a summary of team discussion on this matter, with @delgod involved.

Some of our products (like PGbouncer) are supporting 20.04. We must respect backwards compatibility with applications using our products.

marceloneppel (approved):
LGTM! Just remember to bump `LIBPATCH`, as it's already 32 on the [main](https://github.com/canonical/data-platform-libs/blob/3b06c74e3496e7487acb8974c18ae51196328fb5/lib/charms/data_platform_libs/v0/data_interfaces.py#L334) branch.

lib/charms/data_platform_libs/v0/data_interfaces.py
marceloneppel: Thanks! :) I tested the changes on Indico, and its build worked now.

marceloneppel (approved):

PR #161, opened 2024-04-18
[MISC] Fix Kafka spelling

deusebio:
I have noticed that we have some spelling checks that do not follow the convention we have for databases.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2834,7 +2834,7 @@ class DatabaseRequires(DatabaseRequirerData, DatabaseRequirerEventHandlers):
 # Kafka Events
 
 
-class KafkaProvidesEvent(RelationEvent):
+class KafkaProviderEvent(RelationEvent):
     """Base class for Kafka events."""
 
     @property
@@ -2854,7 +2854,7 @@ class KafkaProvidesEvent(RelationEvent):
         return self.relation.data[self.relation.app].get("consumer-group-prefix")
 
 
-class TopicRequestedEvent(KafkaProvidesEvent, ExtraRoleEvent):
+class TopicRequestedEvent(KafkaProviderEvent, ExtraRoleEvent):
     """Event emitted when a new topic is requested for use on this relation."""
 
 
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2834,7 +2834,7 @@ def __init__(
 # Kafka Events
 
 
-class KafkaProvidesEvent(RelationEvent):
+class KafkaProviderEvent(RelationEvent):
```
deusebio: I have reverted this change, as this would not have been consistent with the Database classes. But I'm wondering why we kept `Provides` here, and not change it to "Provider"? It seems we did not do this for:

* `*ProvidesEvent`
* `*ProvidesEvents`

Are these used at all in any charm and we need to keep backward compatibility on these?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -3206,12 +3206,12 @@ def set_version(self, relation_id: int, version: str) -> None:
         self.update_relation_data(relation_id, {"version": version})
 
 
-class OpenSearchProvidesEventHandlers(EventHandlers):
+class OpenSearchProviderEventHandlers(EventHandlers):
```
juditnovak, line 3209: Was there an agreement on Opensearch side?

This means they (us) need to update the code.

Otherwise they (us) can't update libs.

juditnovak (changes requested):
This is a breaking change for Opensearch. We need to sync with the team.

lib/charms/data_platform_libs/v0/data_interfaces.py
deusebio: Are you already using the new pattern on opensearch? If not, it should be transparent, since this is a new class appearing some PRs ago. I believe this misspelling slipped in the review process (I would regard this as a tiny bug), but up to them (you) what to do. I can roll back. Just be aware you are a tiny bit off the standard there.

welpaolo (approved):
Looks good to me

deusebio:
I have reverted the OS modification. If you'll want to implement that, you can easily one the commit in this PR

juditnovak (approved):

marceloneppel:
The LIBPATCH wasn't bumped.

deusebio:
Yes, we are aware! https://github.com/canonical/data-platform-libs/pull/165

PR #162, opened 2024-04-22
Add split pattern test

zmraul:
- Add the split pattern deployment to the tests on Kafka.
- Fix issue related to split pattern data access

PR #163, opened 2024-04-22
Fix relation data access

zmraul:
Copying from a comment:

Passing `self` to `getattr` means that none of those fields on the array exist on the inheritance tree from `KafkaRequiresEventHandlers`. Because of this, the lib would always write an empty string to databag, so `topic_requested` is never triggered.

Using the full `KafkaRequires` class works because we pass data indirectly to via `KafkaRequiresEventHandlers.__init__(self, charm, self)` The 3rd argument is a self reference, so `KafkaRequiresData` fields are available down the line.

NOTE: alternative, less intrusive fix
```python
relation_data = {
    f: getattr(self.relation_data, f.replace("-", "_"), "")
    for f in ["consumer-group-prefix", "extra-user-roles", "topic"]
}
```

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 34
+LIBPATCH = 35
 
 PYDEPS = ["ops>=2.0.0"]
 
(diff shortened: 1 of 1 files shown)


welpaolo (commented):
Looks good to me!

Can you please copy the test that you have made in this [PR](https://github.com/canonical/data-platform-libs/pull/162/) in the following one?

deusebio (commented):
+1 @welpaolo 

Also, I suppose we should update the names of the classes based on #161

zmraul:
@welpaolo This branch is being merged and based on the one with the test. So it's already passing here :)

deusebio:
Yes, but probably when rebasing you will have conflicts, given we are touching the same lines. Just remember to keep the spelling consistent! :D 

Anyhow, the changes proposed in the PR, looks good to me!

welpaolo (approved):
LGTM

PR #165, opened 2024-04-23
[MISC] Bumping LIBPATCH

deusebio:
@taurus-forever spotted that we did not update the libpatch in the last PR :angry:  . Thanks Alex!

I'm wondering how often this happens, but it seems something that it can easily be overlooked. We should probably have a check to prevent this from happening again. Also, this feels something that can be generally needed by everyone publishing charmlib

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 34
+LIBPATCH = 35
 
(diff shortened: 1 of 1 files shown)


taurus-forever:
We normally have it (e.g. [here](https://github.com/canonical/mysql-operator/blob/main/.github/workflows/ci.yaml#L37)), but: 
1) it is missing for this repo
2) [it doesn't help](https://github.com/canonical/charming-actions/issues/90)

taurus-forever (approved):

zmraul:
@deusebio Already bumping here: https://github.com/canonical/data-platform-libs/pull/162

deusebio:
I'm fine either way depending on which PR lands first and whether this is blocking someone.

@juditnovak feel free to either merge or close this PR as you feel more appropriate

Issue #164, opened 2024-04-22
FEATURE REQUEST - No-op secret.set_content() if the new content is the same as the current content


marcoppenheimer:
## Context
Full context can be found in [this Matrix thread](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$PuD_JzBT2dgSjLkNQr6r6oIVtRdxW03o8-GsG4qbAUM?via=ubuntu.com&via=matrix.org&via=mx.aouss.it)

In summary, IS raised an issue of too many secret revisions causing a slowdown of a production Juju MongoDB snap, causing an outage.
After debugging, the cause was the Kafka charm using DP libs calling the following every `update-status`:
```python
self.kafka_provider.set_tls(relation_id=relation.id, tls=tls)
```

After discussion with the Juju team, it appears that every time we `secret.set_content()`, it creates a new revision, regardless of whether or not the contents have actually changed.

### Issue
Too many secret revisions causes issues for the Juju MongoDB snap.
Currently, there are no protections in `ops` or DP Libs to avoid repeatedly updating a secret field with the same content, which is a fairly reasonable use-case, present in 'Config' and 'Relation Data' setting.
Secrets should work the same way as charm 'Config' and 'Relation Data', where setting the same value ends up in a no-op, doing 'nothing'.
The Juju team said that this won't be possible to implement on their end.

### Request
Make DP Libs check secret content during a desired update to a secret field, and if the update is the same as the current content, do nothing.

### Desired Behavior
In pseudo, the main gist would be to support the following:
```python
def update_clients(self):
    for relation in self.model.relations[CLIENT_RELATION_NAME]:
        secret = get_secret(relation.name)  # let's assume this func exists

        secret.set_content({"some-state": self.get_up_to_date_state()})

def _on_update_status(self, event: UpdateStatusEvent):
    self.update_clients()
```

The actual `get_secret` and `set_content` is of course abstracted by the lib. But essentially whenever the lib **would** write a secret, it should check the content and skip it if it's unchanged.

PR #169, opened 2024-05-22
[DPE-4335] Do not trigger secret revision change if the relation content is the same

Batalex:
This PR prevents the data interface abstraction from triggering the emission of a new secret revision if the content is the same. The behavior is unaltered when using the secret API itself from `data_secrets.py`

Fixes https://github.com/canonical/data-platform-libs/issues/164


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1369,7 +1369,10 @@ def _update_relation_secret(
         old_content = secret.get_content()
         full_content = copy.deepcopy(old_content)
         full_content.update(content)
-        secret.set_content(full_content)
+
+        # DPE-4182: do not create new revision if the content stay the same
```
marcoppenheimer, line 1373: **praise:** Great comment.

marcoppenheimer (approved):
Looks great! Happy with it :)

If you can, add this line to the PR description:

```
Fixed https://github.com/canonical/data-platform-libs/issues/164
```

Then when this PR closes, it'll auto-close the issue.

juditnovak:
Seems as a great change with great tests, thank you very much for it!!!! :-)

I'll get after having verified that this is surely all that was needed to do (as some of the workflows may be a bit "tangled up" :sweat_smile: )

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1369,7 +1369,10 @@ def _update_relation_secret(
...
+        if old_content != full_content:
+            secret.set_content(full_content)
```
juditnovak, line 1375: I think this fix should rather go to the "core" of the issue, where we are gathering additional logic applied on top of base `ops` secret operations. Namely to `CachedSecret.set_content()`. 

This way, any further calls to this method would be protected from unnecessary revision changs.

(Though technically, given the current workflows the two solutions behave the same, the latter is safer for future modifications, re-factors, etc.)

tests/integration/test_charm.py
```diff
@@ -218,6 +223,22 @@ async def test_peer_relation_secrets(component, ops_test: OpsTest):
...
+        "set-peer-relation-field",
+        **{"component": component, "field": "secret-field", "value": "blablabla2"},
+    )
+    await action.wait()
+
+    unchanged_secret_revision = await get_secret_revision_by_label(
+        ops_test, f"database-peers.database.{component}", owner
+    )
+    assert changed_secret_revision == unchanged_secret_revision
+
```
juditnovak, line 241: Would it be possible either to add this as a separate test (nothing much but just to copy-paste it after) or to have clear comments indicating the purpose of this section?

What you are doing is VERY GOOD :-)  My only concern is about the readability and extendibility of our tests. What I mean is to avoid the becoming even more "spagetti" than what they are already :sweat_smile: 

Also, normally this feature should equally apply to cross-charm secrets. I guess we should test the behavior on both levels.

I'm also wondering if a simple unittest may be also nice to have (as this feature should be well "unit-test-able"... (It's helpful for future development to be able to quickly verify that expected behaviors are kept intact before running the pipelines.)

lib/charms/data_platform_libs/v0/data_interfaces.py
Batalex: Done in 1fc744cfa7a453129ce684ad5d26f516224a858c

juditnovak:
The test charms contents are dynamically fetched, we shouldn't keep them updated.
(I guess we should rather do a proper cleanup on that at some point :sweat_smile: )

Batalex:
> The test charms contents are dynamically fetched, we shouldn't keep them updated. (I guess we should rather do a proper cleanup on that at some point 😅 )

I can gitignore them if you'd like. I was committing them for consistency intra repo

juditnovak:
@Batalex I see your point... The reason why I'd advise otherwise is because then we need to maintain this consistency from now on. Normally these files are being copied over from `latestt` just for the sake of the tests. If we now add them to the repo, we need to update them on every single change...

(There was a specific reason why I added the initial version for `dummy-database-charm`... I could agree to update that one. But I rather should look into ways to get rid of that copy next time I'm touching the tests :-)  Let's not have more of those copies if avoidable...)

juditnovak:
The PR is all excellent now in terms of code and integration tests... May I just pls ask for a unittest... 
(Seems like a low-hanging fruit, a simple copy of existing secrets-related tests. If it's more than that, then we can skip.)

tests/integration/test_charm.py
Batalex: Unit test added in 234663f!

Batalex:
@juditnovak I reverted all changes to test charms libs :saluting_face:

juditnovak (approved):
LGTM, wonderful, thank you :-)

PR #170, opened 2024-05-24
[DPE-4416] URI exists while re-creating secret with modified label

juditnovak:
# Issue

Bug reported in Jira ticket https://warthogs.atlassian.net/browse/DPE-4416

The root cause was identified as an issue within the logic implemented for rolling upgades. In particular on an occasion where the (peer) secret label may change (as it happened in v34).

Since Juju is not supporting label changes, whenever we need to change a label for a secret, we need to create a new secret (using the previous content) and associate this new secret with the new label.

(The underlying logic is more complicated than it sounds, as we need to maintain backwards compatibility as well, until the first write operation.)

This logic had a bug, revealing on this particular use case:
 - charm upgrade from libs < v34 to libs > v34
 - peer secret changes
 - peer secret changes for a second time, _within_ the same event handler

At this point internal logic following the label switch was not maintained correctly. Namely the `CachedSecret` object was still associated with the label that --in reality-- we already have deteached from.

# Solution

1. 
Cleaning up the corresponding internal field.

(Also, in a mid-term: refactor and cleanup of the `data_interfaces` module.)

2.
`data_interfaces` had tests to verify that any new version is compatible with a charm using databag.
However we were missing to have a test module that verifies rolling upgrades from previous versions.

A new test module was added, that simulating upgrades from an arbitraty version.
This test is executed in 4 different pipelines, going back from the `latest` version of the lib down to `latest - 4` simulating an upgrade from that version both for peer and cross-charm relations.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 35
+LIBPATCH = 36
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -642,8 +642,8 @@ class CachedSecret:
             return
 
         # Create a new secret with the new label
-        old_meta = self._secret_meta
         content = self._secret_meta.get_content()
+        self._secret_uri = None
 
         # I wish we could just check if we are the owners of the secret...
         try:
@@ -651,7 +651,7 @@ class CachedSecret:
         except ModelError as err:
             if "this unit is not the leader" not in str(err):
                 raise
-        old_meta.remove_all_revisions()
+        self.current_label = None
 
     def set_content(self, content: Dict[str, str]) -> None:
         """Setting cached secret content."""
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -1586,7 +1586,7 @@ class RequirerData(Data):
         """
         label = self._generate_secret_label(relation_name, relation_id, group)
 
-        # Fetchin the Secret's meta information ensuring that it's locally getting registered with
+        # Fetching the Secret's meta information ensuring that it's locally getting registered with
         CachedSecret(self._model, self.component, label, secret_id).meta
 
     def _register_secrets_to_relation(self, relation: Relation, params_name_list: List[str]):
@@ -2309,7 +2309,7 @@ class RelationEventWithSecret(RelationEvent):
         return self._cached_secrets
 
     def _get_secret(self, group) -> Optional[Dict[str, str]]:
-        """Retrieveing secrets."""
+        """Retrieving secrets."""
         if not self.app:
             return
         if not self._secrets.get(group):
--- a/tests/integration/test_rolling_upgrade_from_specific_version.py
+++ b/tests/integration/test_rolling_upgrade_from_specific_version.py
@@ -102,7 +102,6 @@ async def test_deploy_charms(ops_test: OpsTest, application_charm, database_char
 @pytest.mark.parametrize("component", ["app", "unit"])
 async def test_peer_relation(component, ops_test: OpsTest):
     """Peer relation safe across upgrades."""
-    await upgrade_to_new_version(ops_test, DATABASE_APP_NAME)
     await downgrade_to_old_version(ops_test, DATABASE_APP_NAME)
 
     # Setting and verifying two fields (one that should be a secret, one plain text)
@@ -211,9 +210,9 @@ async def test_unbalanced_versions_req_old_vs_prov_new(
     await ops_test.model.wait_for_idle(apps=APP_NAMES, status="active")
 
     # Username
-    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
-    leader_name = f"{APPLICATION_APP_NAME}/{leader_id}"
-    action = await ops_test.model.units.get(leader_name).run_action(
+    leader_app_id = await get_leader_id(ops_test, APPLICATION_APP_NAME)
+    leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
+    action = await ops_test.model.units.get(leader_app_name).run_action(
         "get-relation-field",
         **{"relation_id": pytest.first_database_relation.id, "field": "username"},
     )
@@ -223,9 +222,7 @@ async def test_unbalanced_versions_req_old_vs_prov_new(
     username = action.results.get("value")
 
     # Password
-    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
-    leader_name = f"{APPLICATION_APP_NAME}/{leader_id}"
-    action = await ops_test.model.units.get(leader_name).run_action(
+    action = await ops_test.model.units.get(leader_app_name).run_action(
         "get-relation-field",
         **{"relation_id": pytest.first_database_relation.id, "field": "password"},
     )
@@ -245,8 +242,6 @@ async def test_unbalanced_versions_req_old_vs_prov_new(
     assert action.results.get("value") == username
 
     # Password is correct
-    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
-    leader_name = f"{DATABASE_APP_NAME}/{leader_id}"
     action = await ops_test.model.units.get(leader_name).run_action(
         "get-relation-self-side-field",
         **{"relation_id": pytest.first_database_relation.id, "field": "password"},
@@ -261,7 +256,7 @@ async def test_rolling_upgrade_requires_old_vs_provides_new(ops_test: OpsTest):
     await upgrade_to_new_version(ops_test, APPLICATION_APP_NAME)
 
     # Application charm leader
-    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_id = await get_leader_id(ops_test, APPLICATION_APP_NAME)
     leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
 
     # DB leader
@@ -316,7 +311,7 @@ async def test_rolling_upgrade_requires_old_vs_provides_new_upgrade_new_secret(
 ):
     """After Requires upgrade new secrets are possible to define."""
     # Application charm leader
-    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_id = await get_leader_id(ops_test, APPLICATION_APP_NAME)
     leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
 
     # DB leader
@@ -409,8 +404,6 @@ async def test_unbalanced_versions_req_new_vs_prov_old(
     assert action.results.get("value")
 
     # Password exists
-    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
-    leader_name = f"{DATABASE_APP_NAME}/{leader_id}"
     action = await ops_test.model.units.get(leader_name).run_action(
         "get-relation-self-side-field",
         **{"relation_id": pytest.second_database_relation.id, "field": "password"},
@@ -427,7 +420,7 @@ async def test_rolling_upgrade_requires_new_vs_provides_old(
     await upgrade_to_new_version(ops_test, DATABASE_APP_NAME)
 
     # Application charm leader
-    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_id = await get_leader_id(ops_test, APPLICATION_APP_NAME)
     leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
 
     # DB leader
@@ -482,7 +475,7 @@ async def test_rolling_upgrade_requires_new_vs_provides_old_upgrade_new_secret(
 ):
     """After Provider upgrade, we can safely define new secrets."""
     # Application charm leader
-    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_id = await get_leader_id(ops_test, APPLICATION_APP_NAME)
     leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
 
     # DB leader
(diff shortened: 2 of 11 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -642,16 +642,16 @@ def _move_to_new_label_if_needed(self):
...
         content = self._secret_meta.get_content()
+        self._secret_uri = None
 
         # I wish we could just check if we are the owners of the secret...
         try:
             self._secret_meta = self.add_secret(content, label=self.label)
         except ModelError as err:
             if "this unit is not the leader" not in str(err):
                 raise
-        old_meta.remove_all_revisions()
```
juditnovak, line 654: We need this in a long term, but I rather leave "garbage" around for now, and bring this code back after having verified that under no circumstances we may remove this data pre-mature.

See issue on this matter https://github.com/canonical/data-platform-libs/issues/171

tox.ini
```diff
@@ -100,6 +103,25 @@ deps =
...
+commands =
+    cp tests/integration/data/data_interfaces.py.1 tests/integration/data/data_interfaces.py_old
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade_from_specific_version.py
+    cp tests/integration/data/data_interfaces.py.2 tests/integration/data/data_interfaces.py_old
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade_from_specific_version.py
+    cp tests/integration/data/data_interfaces.py.3 tests/integration/data/data_interfaces.py_old
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade_from_specific_version.py
+    cp tests/integration/data/data_interfaces.py.4 tests/integration/data/data_interfaces.py_old
+    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade_from_specific_version.py
+
```
juditnovak, line 124: Instead of the current hard copy of the previous lib versions, we'll be fetching these files from the git repo dynamically.
The current solution is temporary.

tests/unit/test_data_interfaces.py
```diff
@@ -632,8 +632,6 @@ def test_peer_relation_interface_backwards_compatible_legacy_label(self, interfa
         secret2 = self.harness.model.get_secret(label=f"{PEER_RELATION_NAME}.database.{scope}")
         assert secret2.id != secret_id
         assert interface.fetch_my_relation_field(relation_id, "secret-field") == "blabla"
-        with pytest.raises(SecretNotFoundError):
-            secret = self.harness.model.get_secret(label=f"database.{scope}")
```
juditnovak, line 636: Disabled for now, to be re-enabled in https://github.com/canonical/data-platform-libs/issues/171

taurus-forever (approved):
I am giving LGTM to library changes here, as I have tested them on weekend and the original issue is no longer reproducible. Well done.

Re: tests... I simply cannot load 4*3.5K tests into my mind... 
and this makes me crazy:
```
    cp tests/integration/data/data_interfaces.py.4 tests/integration/data/data_interfaces.py_old
    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade_from_specific_version.py
```

Why do you commit files py.1-4 and copying them into py_old is unclear for me.... please add to PR description the black magic logic. Tnx!

Anyway LGTM for the library fix to unblock PG promotion to candidate for Landscape Team.

delgod (changes requested):
The ticket number should be part of the PR description.
The PR description should clearly explain the problem and how the new code solves it.
At least one technical reviewer should be added.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -642,16 +642,16 @@ def _move_to_new_label_if_needed(self):
             return
 
         # Create a new secret with the new label
-        old_meta = self._secret_meta
         content = self._secret_meta.get_content()
+        self._secret_uri = None
```
juditnovak, line 646: Underlying logic: Normally this helper field is to hold the URI of a secret (in case the secret was newly created, or fetched by URI).

At this point, we are switching the `self` (CachedSecret) object to point to a new secret object (that's associated with the new label).

Thus we have to "unlink" the object from the old URI. (Otherwise new secret creation is blocked: as the error in https://warthogs.atlassian.net/browse/DPE-4416 was indicating so.)

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -642,16 +642,16 @@ def _move_to_new_label_if_needed(self):
...
 
         # I wish we could just check if we are the owners of the secret...
         try:
             self._secret_meta = self.add_secret(content, label=self.label)
         except ModelError as err:
             if "this unit is not the leader" not in str(err):
                 raise
-        old_meta.remove_all_revisions()
+        self.current_label = None
```
juditnovak, line 654: This field is used to indicate when an upgrade was performed, where a secret label change may have happened. 

We detect dynamically if a secret with an old label may be "hanging around", from an old version of the charm. In order to ensure smooth upgrades, we are leaving the secret associated with the old label, as long as only read operations are performed on the secret.
We hold the value of the outdated label in the `CachedSecret.current_label` field (while `CachedSecret.label` is always pointing to the label that is to be used in a long term.

When the above code is executed, we are at the point of the first write operation impacting this secret.
I.e. move to the new label (by having created a brand-new secret, that's recognized by the new label).
We need to clean the `current_label` field, as we have no link to the old (labelled) secret anymore.

delgod (approved):

juditnovak:
@taurus-forever to answer your question.

We have a new test module, that performs a simulated rolling upgrade on test charms.

The initial PR (to deliver a quick yet reliable solution) took a local copy of older versions of the lib, and "rotated" them to a static location for the the test module to pick up. 

As I got the opportunity to clean up the "brute-force" approach, now the tests are dynamically fetching corresponding `latest - N` versions from git on each run of the corresponding, new pipelines.

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+
+    # Storing Relation object in 'pytest' global namespace for the session
+    pytest.first_database_relation = await ops_test.model.add_relation(
+        f"{APPLICATION_APP_NAME}:{FIRST_DATABASE_RELATION_NAME}", DATABASE_APP_NAME
+    )
+    await ops_test.model.wait_for_idle(apps=APP_NAMES, status="active")
+
+    # Username
+    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_name = f"{APPLICATION_APP_NAME}/{leader_id}"
```
marceloneppel, line 215: Shouldn't you use either `DATABASE_APP_NAME` or `APPLICATION_APP_NAME` in both statements? The same happens on the code block after `# Password`.

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+@pytest.mark.abort_on_fail
+async def test_rolling_upgrade_requires_old_vs_provides_new(ops_test: OpsTest):
+    """Upgrading Requires to the latest version of the libs."""
+    await upgrade_to_new_version(ops_test, APPLICATION_APP_NAME)
+
+    # Application charm leader
+    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
```
marceloneppel, line 265: Shouldn't you use `APPLICATION_APP_NAME` in both statements?

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+@pytest.mark.abort_on_fail
+async def test_rolling_upgrade_requires_old_vs_provides_new_upgrade_new_secret(
+    ops_test: OpsTest,
+):
+    """After Requires upgrade new secrets are possible to define."""
+    # Application charm leader
+    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
```
marceloneppel, line 320: Shouldn't you use `APPLICATION_APP_NAME` in both statements?

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+@pytest.mark.abort_on_fail
+async def test_rolling_upgrade_requires_new_vs_provides_old_upgrade_new_secret(
+    ops_test: OpsTest,
+):
+    """After Provider upgrade, we can safely define new secrets."""
+    # Application charm leader
+    leader_app_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_app_name = f"{APPLICATION_APP_NAME}/{leader_app_id}"
```
marceloneppel, line 486: Shouldn't you use `APPLICATION_APP_NAME` in both statements?

tests/integration/test_rolling_upgrade_from_specific_version.py
marceloneppel: As these tests are running on K8s and you have only one unit of each app, you can always consider `0` the leader ID.

tests/unit/test_upgrade.py
```diff
@@ -261,7 +261,7 @@ def test_dependency_model_raises_for_bad_dependency(value):
 def test_dependency_model_raises_for_bad_nested_dependency(value):
     deps = {
         "gandalf_the_white": {
-            "dependencies": {"gandalf_the_grey": "~1.0", "durin": value},
+            "dependencies": {"gandalf_the_grey": "~1.0", "bilbo": value},
```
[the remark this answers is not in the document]
juditnovak: Codespell inisisted that `durin` should be `during`

tox.ini
```diff
@@ -24,6 +24,11 @@ pass_env =
     MODEL_SETTINGS
     LIBJUJU_VERSION_SPECIFIER
     PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION
+allowlist_externals = 
+    cp
+    grep
+    echo
+    bash
```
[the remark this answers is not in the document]
juditnovak: Good point, cleaing it up. (Traces of earlier attempts).

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+# -----------------------------------------------------
+# Testing 'Peer Relation'
+# -----------------------------------------------------
+
+
+@pytest.mark.abort_on_fail
+@pytest.mark.parametrize("component", ["app", "unit"])
+async def test_peer_relation(component, ops_test: OpsTest):
+    """Peer relation safe across upgrades."""
+    await upgrade_to_new_version(ops_test, DATABASE_APP_NAME)
```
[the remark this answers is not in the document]
juditnovak: Indeed this is not needed (though harmless). Removing it.

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+    action = await ops_test.model.units.get(unit_name).run_action(
+        "get-peer-relation-field", **{"component": component, "field": "not-a-secret"}
+    )
+    await action.wait()
+    assert action.results.get("value") == "plain text"
+
+    # Upgrade
+    await upgrade_to_new_version(ops_test, DATABASE_APP_NAME)
+
+    # Both secret and databag content can be modified -- even twice ;-)
```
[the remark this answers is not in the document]
juditnovak: Running it in a loop is a different thing as it results in multiple event handlers.

We need to test in the same event handler (typically for this bug, but also useful in general).

tests/integration/test_rolling_upgrade_from_specific_version.py
```diff
@@ -0,0 +1,545 @@
...
+    action = await ops_test.model.units.get(leader_name).run_action(
+        "get-relation-self-side-field",
+        **{"relation_id": pytest.first_database_relation.id, "field": "username"},
+    )
+    await action.wait()
+    assert action.results.get("value") == username
+
+    # Password is correct
+    leader_id = await get_leader_id(ops_test, DATABASE_APP_NAME)
+    leader_name = f"{DATABASE_APP_NAME}/{leader_id}"
```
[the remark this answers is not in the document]
juditnovak: :facepalm: 

Thanks, good call :sweat_smile:

tox.ini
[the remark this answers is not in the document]
juditnovak: Done

tests/integration/test_rolling_upgrade_from_specific_version.py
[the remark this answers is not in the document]
juditnovak: Done

tests/integration/test_rolling_upgrade_from_specific_version.py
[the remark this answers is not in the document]
juditnovak: Done

tests/integration/test_rolling_upgrade_from_specific_version.py
juditnovak: Done.

Yes, I see your point for k8s. (Thank for the info, I didn't know. Opensearch is VM only :-) ). 

For now I left it consistent, with the "usual" terminology. (What if we start testing the libs for VM too one day.../ :-) )

tests/integration/test_rolling_upgrade_from_specific_version.py
juditnovak: Done.

tests/integration/test_rolling_upgrade_from_specific_version.py
juditnovak: Done.

tests/integration/test_rolling_upgrade_from_specific_version.py
juditnovak: Done.

marceloneppel (approved):
LGTM!

PR #176, opened 2024-06-10
[DPE-4598] Handle upgrade of top of the stack Juju leader

marceloneppel:
## Issue:

The PostgreSQL VM charm rollback test started to fail by the combination of adding TimescaleDB to the Charmed PostgreSQL Snap + updating the `upgrade` library to LIBPATCH 16 (which removed a `defer` call that was hiding the issue fixed by this PR).

Explanation of the issue (considering the tests from the PostgreSQL VM charm):

* In the `test_upgrade_from_edge` test:
    * TimescaleDB in the snap is leading to a longer snap refresh (only some more seconds).
    * With that, Patroni is performing a failover, so a new unit, different from the Juju leader unit (the former primary), becomes the primary.
    * Also with that, the former primary unit, the Juju leader unit, becomes a replica (not a synchronous standby).

* In the `test_fail_and_rollback` test:
    * After the failed upgrade, the `pre-upgrade-check` is run, which sets the `recovery` state for the Juju leader unit (which is a replica).
    * When the refresh is started, the upgrade charm event is fired in each of the three units, setting the state to ready in all of them.
    * The upgrade changed event is fired only in the non-primary units because only the relation data from the Juju leader unit changed (the state changed from `recovery` to `ready`, while in the other units, it was already set to `ready`, so no change, no events firing for the Juju leader unit).
    * As the Juju leader unit is the replica, it’s the top of the stack unit (the one that should be upgraded first). However, as the upgrade relation changed, the event hasn’t fired for it; it cannot start the upgrade (in this case, it is a rollback).

For Kafka and MySQL operators on VM, it hasn’t happened because both always set the upgrade stack with the Juju leader unit as the last unit to be upgraded. For the following codes, remember that the `build_upgrade_stack` is always executed in the Juju leader unit.
* [Kafka code](https://github.com/canonical/kafka-operator/blob/af5fd600dd6bb6767044868df1e8d8f78a3d4b68/src/events/upgrade.py#L77)
* [MySQL code](https://github.com/canonical/mysql-operator/blob/14c06ff88c4e564cd6d098aa213bd03e78e84b52/src/upgrade.py#L97)

## Solution

If the Juju leader unit is the top unit of upgrading the stack, execute the `on_upgrade_changed` handler logic during a rollback. This way, it can run the logic that starts the upgrade process, in the end of the `on_upgrade_changed` handler.

This fix was tested on the following charms to avoid breaking them:
* [Kafka VM charm](https://github.com/canonical/kafka-operator/actions/runs/9471398513?pr=204)
* [MySQL VM charm](https://github.com/canonical/mysql-operator/actions/runs/9471374322?pr=461)

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 16
+LIBPATCH = 17
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
@@ -906,6 +906,17 @@ class DataUpgrade(Object, ABC):
                 except VersionError as e:  # not ready if not passed check
                     logger.error(e)
                     self.set_unit_failed()
+                    return
+                top_unit_id = self.upgrade_stack[-1]
+                top_unit = self.charm.model.get_unit(f"{self.charm.app.name}/{top_unit_id}")
+                if (
+                    top_unit == self.charm.unit
+                    and self.peer_relation.data[self.charm.unit].get("state") == "recovery"
+                ):
+                    # While in a rollback and the Juju leader unit is the top unit in the upgrade stack, emit the event
+                    # for this unit to start the rollback.
+                    self.peer_relation.data[self.charm.unit].update({"state": "ready"})
+                    self.on_upgrade_changed(event)
                     return
             self.charm.unit.status = WaitingStatus("other units upgrading first...")
             self.peer_relation.data[self.charm.unit].update({"state": "ready"})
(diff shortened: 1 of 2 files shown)


paulomach (approved):
Tks for the context. Looks good

lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -907,6 +907,17 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
                     logger.error(e)
                     self.set_unit_failed()
                     return
+                top_unit_id = self.upgrade_stack[-1]
+                top_unit = self.charm.model.get_unit(f"{self.charm.app.name}/{top_unit_id}")
+                if (
+                    top_unit == self.charm.unit
+                    and self.peer_relation.data[self.charm.unit].get("state") == "recovery"
```
paulomach, line 914: iirc, recovery is set by `pre-upgrade-check` in the context of the rollback, so this is reliably set for the rollback case, correct?

taurus-forever (approved):
Nice! CC: @marcoppenheimer JFYI (as you are mentioned in PR description).
Tnx for the deep troubleshooting and the great PR description Marcelo!

PR #179, opened 2024-06-28
[MISC] Don't use planned units for single unit cluster upgrade

dragomirp:
Planned units are not correct for subordinate charms [juju bug](https://bugs.launchpad.net/juju/+bug/2025676), so upgrade will not start on a single unit subordinate charm.

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 17
+LIBPATCH = 18
 
 PYDEPS = ["pydantic>=1.10,<2", "poetry-core"]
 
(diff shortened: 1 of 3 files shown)


lib/charms/data_platform_libs/v0/upgrade.py
```diff
@@ -921,7 +921,7 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
             self.charm.unit.status = WaitingStatus("other units upgrading first...")
             self.peer_relation.data[self.charm.unit].update({"state": "ready"})
 
-            if self.charm.app.planned_units() == 1:
+            if len(self.app_units) == 1:
```
dragomirp, line 924: Alternatively, I can extract this in a separate method, so that subordinates can override it, while principal charms can keep on using the same logic.

dragomirp:
CI should get fixed once  https://github.com/canonical/charm-logdump-action/pull/9 gets merged.

marceloneppel (approved):
LGTM!

I executed the following tests locally with both 1 and 3 units. All passed.
* Kafka VM charm:
  * `tests/integration/test_upgrade.py` 
* MySQL VM charm:
  * `tests/integration/high_availability/test_upgrade.py`
  * `tests/integration/high_availability/test_upgrade_from_stable.py`
* PostgreSQL VM charm:
  * `tests/integration/ha_tests/test_upgrade.py`
  * `tests/integration/ha_tests/test_upgrade_from_stable.py`

lib/charms/data_platform_libs/v0/upgrade.py
paulomach: if it's robust for both cases, keep it simple

paulomach (approved):

PR #180, opened 2024-07-03
[DPE-4772] Subordinate scale up

dragomirp:
Scaling up subordinate charms with a single endpoint is error prone by default, since the endpoint is local, but a new unit may not yet be ready to serve when the new consumer receives a `database_created` event. The PR adds a check to the unit relation databag for a `state` value that can be used to synchronise the emission of events and `subordinated` flag in the provider application databag. The change should be backwards compatible.

PR https://github.com/canonical/pgbouncer-operator/pull/274 implements the other side of the synchronisation.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2842,6 +2842,22 @@ class DatabaseRequirerEventHandlers(RequirerEventHandlers):
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the database relation has changed."""
+        remote_units = [
+            key
+            for key in event.relation.data.keys()
+            if isinstance(key, Unit) and key.name.startswith(event.relation.app.name)
+        ]
+        # Check that provider units have joined.
+        if len(remote_units) == 0:
+            logger.debug("No provider units are available.")
+            return
+        elif (
+            len(remote_units) == 1
+            and event.relation.data[remote_units[0]].get("state", "ready") != "ready"
+        ):
+            logger.debug("Subordinate provider unit not ready.")
+            return
+
         # Check which data has changed to emit customs events.
         diff = self._diff(event)
 
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 37
+LIBPATCH = 38
 
 PYDEPS = ["ops>=2.0.0"]
 
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2855,7 +2855,7 @@ class DatabaseRequirerEventHandlers(RequirerEventHandlers):
         is_subordinate = False
         remote_unit_data = None
         for key in event.relation.data.keys():
-            if isinstance(key, Unit) and key.name.startswith(self.charm.app.name):
+            if isinstance(key, Unit) and not key.name.startswith(self.charm.app.name):
                 remote_unit_data = event.relation.data[key]
             elif isinstance(key, Application) and key.name != self.charm.app.name:
                 is_subordinate = event.relation.data[key].get("subordinated") == "true"
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2606,6 +2606,16 @@ class DatabaseProviderData(ProviderData):
         """
         self.update_relation_data(relation_id, {"version": version})
 
+    def set_subordinated(self, relation_id: int) -> None:
+        """Raises the subordinated flag in the application relation databag.
+
+        The flag will be used to evaluate additional checks before emitting provider events.
+
+        Args:
+            relation_id: the identifier for a particular relation.
+        """
+        self.update_relation_data(relation_id, {"subordinated": "true"})
+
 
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2842,22 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
...
+            return
+        elif (
+            len(remote_units) == 1
+            and event.relation.data[remote_units[0]].get("state", "ready") != "ready"
+        ):
+            logger.debug("Subordinate provider unit not ready.")
+            return
+
         # Check which data has changed to emit customs events.
         diff = self._diff(event)
```
dragomirp, line 2862: We should do the subordinate checks before the diff is updated, otherwise custom event behaviour will change.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2842,22 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
...
+            key
+            for key in event.relation.data.keys()
+            if isinstance(key, Unit) and key.name.startswith(event.relation.app.name)
+        ]
+        # Check that provider units have joined.
+        if len(remote_units) == 0:
+            logger.debug("No provider units are available.")
+            return
+        elif (
+            len(remote_units) == 1
```
dragomirp, line 2855: Subordinate relations will have a single remote unit, the one subordinated in the local unit. If we have multiple remote units, we don't have a subordinate relation.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2842,22 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
...
+            and event.relation.data[remote_units[0]].get("state", "ready") != "ready"
```
dragomirp, line 2856: Defaults to ready, so that existing behaviour doesn't change. The subordinate provider should set a different `state` in the `relation_created` hook.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2606,6 +2606,16 @@ def set_version(self, relation_id: int, version: str) -> None:
...
+    def set_subordinated(self, relation_id: int) -> None:
+        """Raises the subordinated flag in the application relation databag.
+
+        The flag will be used to evaluate additional checks before emitting provider events.
+
+        Args:
+            relation_id: the identifier for a particular relation.
+        """
+        self.update_relation_data(relation_id, {"subordinated": "true"})
+
```
dragomirp, line 2618: Helper to set the application level flag. If set the Requirer will do the additional subordination checks.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2852,24 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the database relation has changed."""
+        is_subordinate = False
+        remote_unit_data = None
+        for key in event.relation.data.keys():
+            if isinstance(key, Unit) and key.name.startswith(self.charm.app.name):
+                remote_unit_data = event.relation.data[key]
```
dragomirp, line 2859: Should only have one remote unit in case of subordinated relation

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2852,24 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         """Event emitted when the database relation has changed."""
+        is_subordinate = False
+        remote_unit_data = None
+        for key in event.relation.data.keys():
+            if isinstance(key, Unit) and key.name.startswith(self.charm.app.name):
```
marceloneppel, line 2858: Fix to get the remote unit data:
```suggestion
            if isinstance(key, Unit) and not key.name.startswith(self.charm.app.name):
```

marceloneppel (changes requested):

marceloneppel (approved):
Looks great! Thanks @dragomirp!

delgod (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2842,6 +2852,24 @@ def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
...
+            elif isinstance(key, Application) and key.name != self.charm.app.name:
+                is_subordinate = event.relation.data[key].get("subordinated") == "true"
+
+        if is_subordinate:
+            # Check that provider units have joined.
+            if not remote_unit_data:
+                logger.debug("No provider units are available.")
+                return
+
+            if remote_unit_data.get("state") != "ready":
```
paulomach, line 2869: who/where does `ready` is set?

paulomach (approved):
Goog. Left a question

lib/charms/data_platform_libs/v0/data_interfaces.py
dragomirp: The subordinate provider sets it when the unit is ready.

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: As commented on the related PR: https://github.com/canonical/pgbouncer-operator/pull/274#discussion_r1667025361

This is kind of a "hack" over the concept pf the lib, where cross-charm relations were strictly using `app` data only. That concept must be extended now, instead of direct access to unit/app data in various "random" places within the code. 

I've opened issue https://github.com/canonical/data-platform-libs/issues/182 and I try to address it in the near future (right after mid-cycle)

PR #183, opened 2024-07-08
[MISC] Reduce the size of the library

dragomirp:
Publishing of the data_interfaces lib fails due to size limitations:
```
Store operation failed:
- None: Content too large (got 131339 bytes, limit is 131072 in utf8)
```

Deleting some logs and comments to push below the limit

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2609,8 +2609,6 @@ class DatabaseProviderData(ProviderData):
     def set_subordinated(self, relation_id: int) -> None:
         """Raises the subordinated flag in the application relation databag.
 
-        The flag will be used to evaluate additional checks before emitting provider events.
-
         Args:
             relation_id: the identifier for a particular relation.
         """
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2861,13 +2859,10 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
...
 
         if is_subordinate:
-            # Check that provider units have joined.
             if not remote_unit_data:
-                logger.debug("No provider units are available.")
                 return
 
             if remote_unit_data.get("state") != "ready":
-                logger.debug("Subordinate provider unit not ready.")
                 return
```
dragomirp, line 2866: Going to be refactored anyway.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2609,8 +2609,6 @@ def set_version(self, relation_id: int, version: str) -> None:
     def set_subordinated(self, relation_id: int) -> None:
         """Raises the subordinated flag in the application relation databag.
 
-        The flag will be used to evaluate additional checks before emitting provider events.
-
```
dragomirp, line 2613: Added with the latest change.

dragomirp:
This should give us 14 bytes. We should use them wisely.

marceloneppel (approved):
LGTM!

juditnovak (approved):

juditnovak:
If this is indeed the problem, then I'd take https://warthogs.atlassian.net/browse/DPE-4477 with a high priority.

As it is a blocker

dragomirp:
Merging to see if it unblocks publishing.

Issue #193, opened 2024-08-30
[Missing] data interfaces should provide a property to ensure `database_requested` has run


Gu1nness:
If a charm needs to do some updates of the stored data/secrets in a `relation_changed` event for example, it has to check that the `database_requested` event has run. Currently, there is no documentation on how to do that, nor a property or a check function to do that.

Adding a `is_relation_setup` or equivalent + the according documentation and a small example should do the thing.

juditnovak:
Thank you very much for reporting. Indeed it's a serious concern.

Instead of property + docs what we provide is the following: as long as the `database` field is not set on the relation, the `Provicer` (the side creating secrets) will not be allowed to apply write operation.

If the `Provider` would make an attempt to execute` update_relation_data()` it would throw an exception indicating that the opration is premature.

PR #194, opened 2024-09-08
[DPE-5306][SECURITY] No write before protocol init

juditnovak:
# Issue

https://github.com/canonical/data-platform-libs/issues/193

In case any write operation may take place before the protocol is established the `Provider` may not yet be aware whether secrets are to be used or not. Thus, the write operation may end up as plain text in the databag, instead of secrets.

Example:

A charm may receive TLS certs anytime, that are to be shared on a client relation.
When the charm got the TLS secrest, it just "blindly" writes them using `update_relation_data()` on `update-status` or `certificate-available` or so.
Which technically could be executed BEFORE the  `Requirer` (!!!!) `relation-created` handler would put the `requested-secrets` list -- highlighting that TLS certs are stored as a Juju Secret on this relation.

# Solution

No `Provider` write operation is allowed before the initial part of the protocol is finished. 

This can be assured by waiting until the `Requirer` would add the `database` field to the Relation Data. By this time the `Requirer` must have executed its `relation-created` hook (where `requested-secrets` is written to the databag)

NOTE: Given that the change is straightforward, while the scenario is non-trivial to reproduce. I believe that unittest coverage is sufficient.

NOTE2: This may be a **breaking change** for charms that were publishing sensitive data premature. Clearly corresponding charm code is to be fixed.

See PG corresponding fix with healthy pipelines: https://github.com/canonical/postgresql-operator/pull/615

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -1457,7 +1457,7 @@ class EventHandlers(Object):
 class ProviderData(Data):
     """Base provides-side of the data products relation."""
 
-    DATABASE_FIELD = "database"
+    RESOURCE_FIELD = "database"
 
     def __init__(
         self,
@@ -1626,7 +1626,7 @@ class ProviderData(Data):
         req_secret_fields = []
 
         keys = set(data.keys())
-        if self.fetch_relation_field(relation.id, self.DATABASE_FIELD) is None and (
+        if self.fetch_relation_field(relation.id, self.RESOURCE_FIELD) is None and (
             keys - {"endpoints", "read-only-endpoints", "replset"}
         ):
             raise PrematureDataAccessError(
@@ -3305,7 +3305,7 @@ class KafkaRequiresEvents(CharmEvents):
 class KafkaProviderData(ProviderData):
     """Provider-side of the Kafka relation."""
 
-    DATABASE_FIELD = "topic"
+    RESOURCE_FIELD = "topic"
 
     def __init__(self, model: Model, relation_name: str) -> None:
         super().__init__(model, relation_name)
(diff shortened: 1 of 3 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1618,6 +1620,12 @@ def _fetch_my_specific_relation_data(
     def _update_relation_data(self, relation: Relation, data: Dict[str, str]) -> None:
         """Set values for fields not caring whether it's a secret or not."""
         req_secret_fields = []
+
+        if self.fetch_relation_field(relation.id, self.DATABASE_FIELD) is None:
+            raise DataInterfacesError(
+                "Premature access to relation data, update is forbidden before the connection is initialized."
+            )
+
```
juditnovak, line 1628: Note: As for current conventions, the field MUST exist, but may be empty.

https://github.com/canonical/postgresql-test-app/blob/main/src/charm.py#L155

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1618,6 +1620,12 @@ def _fetch_my_specific_relation_data(
     def _update_relation_data(self, relation: Relation, data: Dict[str, str]) -> None:
         """Set values for fields not caring whether it's a secret or not."""
         req_secret_fields = []
+
+        if self.fetch_relation_field(relation.id, self.DATABASE_FIELD) is None:
```
juditnovak, line 1624: Note: As for current conventions, the field MUST exist, but may be empty.

https://github.com/canonical/postgresql-test-app/blob/main/src/charm.py#L155

delgod (approved):

delgod (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1453,6 +1457,8 @@ def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
 class ProviderData(Data):
     """Base provides-side of the data products relation."""
 
+    DATABASE_FIELD = "database"
```
welpaolo, line 1460: `DATABASE_FIELD` is a little bit misleading maybe we can find a better name such as `RESOURCE_FIELD`

welpaolo (approved):
Look good to me!

marceloneppel (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
juditnovak: Done

PR #196, opened 2024-10-22
[BUGFIX] Fixing delete_relation_data()

juditnovak:
Currently when invoking `delete_relation_data()` with a list of fields to remove, and any of those may not exist (a secret), the whole deletion fails. 

This is not correct. Providing a fix to rather ignore non-existent fields, yet still to go ahead with the deletion of existing ones.


paulomach:
closing stale PRs

PR #198, opened 2024-11-13
Provide abstraction point for setting charm status

addyess:
### Overview

In order to adopt this library into the [`k8s` charm](https://github.com/canonical/k8s-operator) we'd request this abstraction point in the library to not be able to assert full control of the status of the charm.  Setting the charm to `ActiveStatus()` and other states doesn't reflect the needs of our charm users.

### Details
* Adds abstraction methods for setting the status of the charm
* Implements those abstraction methods with the original direct status settings

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -86,6 +86,7 @@ deps =
     juju{env:LIBJUJU_VERSION_SPECIFIER:==2.9.42.4}
     pytest-operator
     pytest-mock
+    websockets<14
     -r {tox_root}/requirements.txt
 commands =
     pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_charm.py
@@ -98,6 +99,7 @@ deps =
     juju{env:LIBJUJU_VERSION_SPECIFIER:==2.9.42.4}
     pytest-operator
     pytest-mock
+    websockets<14
     -r {tox_root}/requirements.txt
 commands =
     pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/test_rolling_upgrade.py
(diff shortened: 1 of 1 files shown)


addyess:
@welpaolo @delgod and @marceloneppel 

I can't retrigger the failed job, but it should work if run again.  

thanks for the review

addyess:
@juditnovak are you the right one to ping regarding PRs here?

welpaolo:
Hi @addyess, I retriggered the failed test. I also added the the more relevant reviewers based on the context of the PR.

marcoppenheimer:
@addyess Thank you for the contribution, and really pleased to hear that other teams are looking at to use our components also!

For context, the motivation behind having a lib manage status, is that during the upgrade of a stateful application, we essentially halt/defer other potentially destructive events while the upgrade is in progress. As a result, during the 'upgrade process' we essentially surrender control of the Status to the lib, as it's likely the most important status to surface.

The changes that you've shared here are interesting, am I correct in assuming that you would override `status_xyz` when instantiating `DataUpgrade`, so that it no-ops?

addyess:
@marcoppenheimer and @carlcsaposs-canonical 

> The changes that you've shared here are interesting, am I correct in assuming that you would override status_xyz when instantiating DataUpgrade, so that it no-ops?

Correct in that i have some concerns handing over full control the the library.  The biggest one I don't like is having the charm ever be in `ActiveStatus()`.   After the upgrade is complete, i want to make sure the charm has a chance to test that post-upgrade there's no issues that would result in another Waiting or Blocked state. 


> would be curious to hear more about your use case—are you doing in-place upgrades of a full Kubernetes cluster? are you supporting rollbacks? is that complexity handled by the charm or the snap? is the charm deployed as a single unit or multiple units?

yes, my use case is the k8s and k8s-worker charms (https://github.com/canonical/k8s-operator) in which we want to only ever upgrade.  Downgrades/Rollbacks in kubernetes aren't really sanctioned.  Heck -- most k8s deployments are suggested to be blue/green.  But this feature is required for things like k8s on metal where blue/green isn't always feasible.  So these are in-place upgrades of the kubernetes cluster via a snap upgrade.

addyess:
We've confirming using the library as it that upgrades are functional without these change

PR #199, opened 2024-11-29
[DPE-5588] Create architecture.py module

sinclert-canonical:
This PR creates a new Python module called `architecture.py` in order to group all shared code related to architecture checking / operating. So far, it only contains code extracted from the PostgreSQL k8s operator (see [code](https://github.com/canonical/postgresql-k8s-operator/blob/d8f8226eeb65cbda25949b043f81a97d7f42a581/src/arch_utils.py)), in order to use it in all MySQL operators without copy-pasting the same ~200 LOC.

### Additional considerations:
- Most of the code proposed was created by @lucasgameiroborges, not myself.
- All integration charms base syntaxes have been migrated to a more modern form, explicitly stating the architecture they are targeting (`amd64` by default).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,6 +58,7 @@ jobs:
           - series: focal
             bases-index: 1
         tox-environments:
+          - integration-arch
           - integration-db
           - integration-upgrade-1
           - integration-upgrade-2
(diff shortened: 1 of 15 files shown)


sinclert-canonical:
Due to the cost of setting up ARM64 runners in this repository, we are going to define the architecture mismatch checks in the MySQL charm lib instead.

PR #200, opened 2024-12-04
Pin websockets to <14.0

sinclert-canonical:
This PR pins `websockets` to use versions smaller than `14.0`, as that version breaks CI (see [runs during the last ~3 weeks](https://github.com/canonical/data-platform-libs/actions/workflows/ci.yaml)).


taurus-forever (approved):
This can be merged to unbliock us, but then can be reverted once new libjuju 2.9 is  released: https://github.com/juju/python-libjuju/issues/1226#issuecomment-2517635734

welpaolo (approved):
LGTM

PR #202, opened 2025-01-14
[MISC] Fix unit tests and add juju 3.6 test runs

dragomirp:
* Fix unit test failures e.g. [here](https://github.com/canonical/data-platform-libs/actions/runs/12759167499/job/35562470069#step:4:559)
* Add juju 3.6 to the integration test matrix

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,6 @@ jobs:
         juju-version:
           - libjuju-version: "2.9.42.4"
           - libjuju-version: "3.2.0.1"
-          - libjuju-version: "3.6.1.0"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
dragomirp, line 1: We should most probably also remove some of the other Juju 3 versions

tests/unit/test_data_secrets.py
```diff
@@ -43,14 +43,14 @@ def test_cached_secret_is_cached(harness, mocker):
     secret = CachedSecret(harness.charm, "mylabel")
     secret.add_secret(content={"rumour": "Community Movie on the way"}, scope="app")
     patched_get = mocker.patch("ops.model.Model.get_secret")
-    patched_get_content = mocker.patch("ops.Secret.get_content")
+    patched_get_content = patched_get.return_value.get_content
```
dragomirp, line 46: `get_content()` is called from the mock `get_secret()`

welpaolo (approved):

.github/workflows/ci.yaml
taurus-forever: @deusebio @delgod @Mehdi-Bendriss do you mind to clean the list here and keep 2.9 and 3.6 LTS only?

LGTM from me to test supported Juju releases only.

taurus-forever (approved):

PR #206, opened 2025-02-26
[DPE-6555] Add read only URIs

dragomirp:
Add a `read-only-uri` secret field. Implements DA158.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 41
+LIBPATCH = 42
 
 PYDEPS = ["ops>=2.0.0"]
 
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1700,7 +1701,7 @@ def set_tls_ca(self, relation_id: int, tls_ca: str) -> None:
 class RequirerData(Data):
     """Requirer-side of the relation."""
 
-    SECRET_FIELDS = ["username", "password", "tls", "tls-ca", "uris"]
+    SECRET_FIELDS = ["username", "password", "tls", "tls-ca", "uris", "read-only-uris"]
```
dragomirp, line 1704: Requirers using an older lib will *not* request this, so for the time being the charm *must* check that the URI will be a secret before setting.

paulomach (approved):
Sane

lib/charms/data_platform_libs/v0/data_interfaces.py
paulomach: I am not sure I understood the issue, can you elaborate?

dragomirp: If a requirer with `data_interfaces` v41 connects to a provider with `data_interfaces` v42, the requirer will send the old SECRET_FIELDS (without `read-only-uris`), if then the provider sets `read-only-uris` it will be put in the relation data and not in the user secret.

So the provider must check what's been written as `requested-secrets` in the relation data:
https://github.com/canonical/postgresql-k8s-operator/blob/3bd2fc0c12c62225f27a59b95e1347485f25b2d3/src/relations/postgresql_provider.py#L228-L233

delgod (approved):

marceloneppel (approved):

taurus-forever (approved):

Gu1nness (approved):

PR #207, opened 2025-02-27
[DPE-6457] - Add etcd_client data interface

skourta:
- Add `EtcdRequires` and `EtcdProvides`
- Refactor `RequirerData` to also be able to use secrets.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 41
+LIBPATCH = 42
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -435,7 +435,7 @@ def set_encoded_field(
 
 
 def diff(event: RelationChangedEvent, bucket: Optional[Union[Unit, Application]]) -> Diff:
-    """Retrieves the diff of the data in the relation changed databag.
+    """Retrieve the diff of the data in the relation changed databag.
 
     Args:
         event: relation changed event.
@@ -500,7 +500,7 @@ def leader_only(f):
 
 
 def juju_secrets_only(f):
-    """Decorator to ensure that certain operations would be only executed on Juju3."""
+    """Ensure that certain operations would be only executed on Juju3."""
 
     def wrapper(self, *args, **kwargs):
         if not self.secrets_enabled:
@@ -539,7 +539,7 @@ def either_static_or_dynamic_secrets(f):
 
 
 def legacy_apply_from_version(version: int) -> Callable:
-    """Decorator to decide whether to apply a legacy function or not.
+    """Decide whether to apply a legacy function or not.
 
     Based on LEGACY_SUPPORT_FROM module variable value, the importer charm may only want
     to apply legacy solutions starting from a specific LIBPATCH.
@@ -585,6 +585,7 @@ class SecretGroupsAggregate(str):
     def __init__(self):
         self.USER = SecretGroup("user")
         self.TLS = SecretGroup("tls")
+        self.MTLS = SecretGroup("mtls")
         self.EXTRA = SecretGroup("extra")
 
     def __setattr__(self, name, value):
@@ -977,6 +978,7 @@ class Data(ABC):
         self.component = self.local_app if self.SCOPE == Scope.APP else self.local_unit
         self.secrets = SecretCache(self._model, self.component)
         self.data_component = None
+        self._secret_fields = list(self.SECRET_LABEL_MAP.keys())
 
     @property
     def relations(self) -> List[Relation]:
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -999,54 +1002,239 @@ def secret_label_map(self):
...
     # Optional overrides
 
     def _legacy_apply_on_fetch(self) -> None:
-        """This function should provide a list of compatibility functions to be applied when fetching (legacy) data."""
+        """Provide a list of compatibility functions to be applied when fetching (legacy) data."""
         pass
 
     def _legacy_apply_on_update(self, fields: List[str]) -> None:
-        """This function should provide a list of compatibility functions to be applied when writing data.
+        """Provide a list of compatibility functions to be applied when writing data.
```
Gu1nness, line 1230: I think the former documentation was better as it tells what should this function (which is an optional override) should do if it is overriden.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -999,54 +1002,239 @@ def secret_label_map(self):
...
 
         Since data may be at a legacy version, migration may be mandatory.
         """
         pass
 
     def _legacy_apply_on_delete(self, fields: List[str]) -> None:
-        """This function should provide a list of compatibility functions to be applied when deleting (legacy) data."""
+        """Provide a list of compatibility functions to be applied when deleting (legacy) data."""
```
Gu1nness, line 1237: I think the former documentation was better as it tells what should this function (which is an optional override) should do if it is overriden.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1304,7 +1491,7 @@ def get_relation(self, relation_name, relation_id) -> Relation:
     def get_secret_uri(self, relation: Relation, group: SecretGroup) -> Optional[str]:
         """Get the secret URI for the corresponding group."""
         secret_field = self._generate_secret_field_name(group)
-        return relation.data[self.component].get(secret_field)
+        return relation.data[relation.app].get(secret_field)
```
Gu1nness, line 1494: This changes the behavior of the library.
This has been designed to support secrets for units and apps, and I think this change breaks this design.
Can you provide reasons why this would not break this design?
Moreover, it does not go in mirror of the `set_secret_uri` below which sets the databag on `self.component`

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1433,6 +1620,46 @@ def __init__(self, charm: CharmBase, relation_data: Data, unique_key: str = ""):
...
+
+        self.framework.observe(
+            self.charm.on[relation_data.relation_name].relation_joined,
+            self._on_relation_joined_event,
+        )
+
+        self.framework.observe(
+            charm.on.secret_changed,
+            self._on_secret_changed_event,
+        )
```
Gu1nness, line 1636: No need to listen to it if you pass/raise on it I think, it will be handled by the subclasses.

tests/unit/test_data_interfaces.py
```diff
@@ -980,15 +982,6 @@ def test_set_additional_fields_secrets(self):
...
 
-        relation_data == {
-            "data": json.dumps(
-                {REQ_SECRET_FIELDS: self.SECRET_FIELDS, self.DATABASE_FIELD: DATABASE}
-            ),  # Data is the diff stored between multiple relation changed events.   # noqa
-            "replset": "rs0",
-            "version": "1.0",
-            "uris": "host1:port,host2:port",
-        }
-
```
Gu1nness, line 991: I think this should stay but with an assert before.

tests/unit/test_data_interfaces.py
```diff
@@ -1021,15 +1014,6 @@ def test_change_additional_fields_secrets(self, secret_changed_mocked):
...
         assert secret_id
 
-        relation_data == {
-            "data": '{REQ_SECRET_FIELDS: "'
-            + f"{json.dumps(self.SECRET_FIELDS)}"
-            + '"}',  # Data is the diff stored between multiple relation changed events.   # noqa
-            "replset": "rs0",
-            "version": "1.0",
-            "uris": "host1:port,host2:port",
-        }
```
Gu1nness, line 1031: I think this should stay but with an assert before.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2368,11 +2282,15 @@ def _delete_relation_data(self, relation: Relation, fields: List[str]) -> None:
                 self.secret_fields,
                 fields,
                 self._update_relation_secret,
-                data={field: self.deleted_label for field in fields},
+                data=dict.fromkeys(fields, self.deleted_label),
```
Gu1nness, line 2285: The former code is faster (not that it matters a lot but I still a win) and more canonical I think

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1700,64 +1748,23 @@ def set_tls_ca(self, relation_id: int, tls_ca: str) -> None:
...
         relation_name: str,
         extra_user_roles: Optional[str] = None,
         additional_secret_fields: Optional[List[str]] = [],
     ):
-        """Manager of base client relations."""
+        """Manage base client relations."""
         super().__init__(model, relation_name)
         self.extra_user_roles = extra_user_roles
-        self._secret_fields = list(self.SECRET_FIELDS)
+        self._secret_fields = list(self.SECRET_LABEL_MAP.keys())
```
Gu1nness, line 1761: That is a great change improvement!

skourta:
> * Can you please provide more information on why such big changes are needed?
> * Can you please ensure that this changes will not break a charm upon upgrade from a former library version to this one?
> * Can you please roll back all changes on docstrings of all methods you did not update? This looks a lot like automated changes and it swarms the PR which is already big with a lot of tiny changes.
> 
> If there are some breaking changes brought by the PR which does not outlive a charm refresh, I think it will be a great opportunity to move to LIBAPI 1.
> 
> I think it will also require huge changes to documentation because the usage in itself changes a bit and also the capabilities of the library.
> 
> Finally, how do you ensure that there is no race condition between the provider and the requirer?
> 
> The current workflow is sound because it enforces an ordering of the sequence of events:
> 
> * requirer relation joined and writes the requested-secrets
> * provider writes the secret-uri
> * requirer gets the data through secret URI and configures its application.
> 
> I'm wondering if your case wouldn't be a bit more fit for two interfaces, one being a send-ca-secret (using [certificate transfer interface](https://charmhub.io/integrations/certificate_transfer/draft) ) + a regular client relation ?

Thank you for the review.
There are a few changes requested for the spec, so I will be working on adding those changes. I will then schedule a call to go through the PR and explain what changes I had to make and the reasons behind them. I think that would answer all of your questions @Gu1nness. 

I think it would be easier for everyone to review the PR after that.

lib/charms/data_platform_libs/v0/data_interfaces.py
skourta: Thank you for the feedback @Gu1nness that is indeed correct. I have patched the function to only use `relation.app` if the secret is not a local secret. Non-local secrets are always set on the app level.

lib/charms/data_platform_libs/v0/data_interfaces.py
skourta: The events are registered here so that the subclass would only need to redefine the event handler.

tests/unit/test_data_interfaces.py
skourta: The test was broken. I have fixed and updated it. I think it was due to the URIs being updated to be a secret rather than a normal field.

tests/unit/test_data_interfaces.py
skourta: Fixed. Another broken test.

lib/charms/data_platform_libs/v0/data_interfaces.py
skourta: This is highly subjective. This was fixed by the linter for not adhering to [C420](https://docs.astral.sh/ruff/rules/unnecessary-dict-comprehension-for-iterable/).

dragomirp (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
Gu1nness: Ok thanks I now understand.
I was a bit lost with all the changes but this was just moved around in the end?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2347,6 +2362,8 @@ def _fetch_my_specific_relation_data(
     @either_static_or_dynamic_secrets
     def _update_relation_data(self, relation: Relation, data: Dict[str, str]) -> None:
         """Update data available (directily or indirectly -- i.e. secrets) from the relation for owner/this_app."""
+        self._load_secrets_from_databag(relation)
```
Gu1nness, line 2365: nit: Pretty sure that it can be great to make a decorator out of this since we add it multiple times as the first step of a call

Gu1nness (commented):
Overall this is excellent work!
I left a few questions and comments, I'm trying to be really picky because this library is important for all DP and I don't want to introduce changes that break in any way.

I think it needs some more tests for the fields you added (unit test and integration tests with a small testing charm), I can give you hints on how to do that.

We should ensure also that we did not go over the lib size limit, which IIRC is 4096 lines? Need to check.

Should we add a custom event like `on_secrets_provided` to react to this exactly ? Handle some secret changed cases maybe?

Finally, there's a big need of documentation update to add this new usecase.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -999,38 +1008,250 @@ def secret_label_map(self):
...
+    # _remote_secret_fields -> PROVIDED_SECRET_FIELDS
+    def _load_secrets_from_databag(self, relation: Relation) -> None:
+        """Load secrets from the databag."""
+        requested_secrets = get_encoded_list(relation, relation.app, REQ_SECRET_FIELDS)
+        provided_secrets = get_encoded_list(relation, relation.app, PROVIDED_SECRET_FIELDS)
+        if requested_secrets is not None:
+            self._secret_fields = requested_secrets
+
+        if provided_secrets is not None:
+            self._remote_secret_fields = provided_secrets
```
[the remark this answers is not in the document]
skourta: We can do that. Currently, it's implemented for the provider and the rest override the implementation.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -999,38 +1008,250 @@ def secret_label_map(self):
...
+            return False
+
+        content = self._content_for_secret_group(data, secret_fields, group_mapping)
+
+        label = self._generate_secret_label(self.relation_name, relation.id, group_mapping)
+        secret = self.secrets.add(label, content, relation)
+
+        # According to lint we may not have a Secret ID
+        if uri_to_databag and secret.meta and secret.meta.id:
+            self.set_secret_uri(relation, group_mapping, secret.meta.id)
```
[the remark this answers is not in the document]
skourta: I don't think the return value is currently used in any way so it is not that critical. I kept the function exactly as it is.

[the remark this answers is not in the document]
skourta: I believe we should keep it as is. The checks are mostly for the linter. We add the secret right before. The main check is the `uri_to_databag`

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -999,38 +1008,250 @@ def secret_label_map(self):
...
+        full_content = copy.deepcopy(old_content)
+        full_content.update(content)
+        secret.set_content(full_content)
+
+        # Return True on success
+        return True
+
+    @juju_secrets_only
+    def _delete_relation_secret(
+        self, relation: Relation, group: SecretGroup, secret_fields: List[str], fields: List[str]
```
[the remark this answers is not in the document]
skourta: I believe it would be better to clean the code in an independent PR. This one has already exploded into a big PR. There are a quite the cleaning work that can be done.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1304,7 +1524,14 @@ def get_relation(self, relation_name, relation_id) -> Relation:
     def get_secret_uri(self, relation: Relation, group: SecretGroup) -> Optional[str]:
         """Get the secret URI for the corresponding group."""
         secret_field = self._generate_secret_field_name(group)
-        return relation.data[self.component].get(secret_field)
+        # if the secret is not managed by this component,
+        # we need to fetch it from the other side
+
+        # Fix for the linter
+        if self.my_secret_groups is None:
+            raise DataInterfacesError("Secrets are not enabled for this component")
```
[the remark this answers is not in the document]
skourta: As per the comment the check is to silence the linter and make the CI pass. if we make the suggestion the linter will complain again.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1712,52 +1783,17 @@ def __init__(
         """Manager of base client relations."""
         super().__init__(model, relation_name)
         self.extra_user_roles = extra_user_roles
-        self._secret_fields = list(self.SECRET_FIELDS)
+        self._secret_fields = self.SECRET_FIELDS
```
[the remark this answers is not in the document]
skourta: Yes that is correct. I do not want to make this PR bigger than it already is.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1700,7 +1771,7 @@ def set_tls_ca(self, relation_id: int, tls_ca: str) -> None:
 class RequirerData(Data):
     """Requirer-side of the relation."""
 
-    SECRET_FIELDS = ["username", "password", "tls", "tls-ca", "uris"]
+    SECRET_FIELDS = ["mtls-chain"]
```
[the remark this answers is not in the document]
skourta: I didn't want to add an additional field that could be calculated. The code has already changed too much for it to stay consistent. I believe splitting the fields would be clearer but introduces an extra unnecessary field.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1466,197 +1714,20 @@ def __init__(
...
-        if relation.app:
-            secret_fields = get_encoded_list(relation, relation.app, REQ_SECRET_FIELDS)
-
-        return self._fetch_relation_data_with_secrets(
-            self.local_app,
-            secret_fields,
-            relation,
-            fields,
-        )
+        self._secret_fields = []
```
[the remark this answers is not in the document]
skourta: Could you please be more specific? This will be set when we load the requested secrets from the requirer. This is nothing but an initialization. The default value in `Data` class is not adequate for the provider.

lib/charms/data_platform_libs/v0/data_interfaces.py
skourta: Sometimes there are additional tests to do before calling the function but I think we can move those tests inside as well. I don't mind either way but slightly prefer the normal call instead of decorators (I dislike decorators, personal preference).

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: I think it's better that each class have their own implementation and this stays as an abstract method.
The reason is that if this gets extended, the developer will have to write their implementation of `_load_secrets_from_databag` instead of getting one that does not necessarily fits their needs.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: If it's purely to please the linter, I think you can add an exception to the linter here instead of adding complexity in the checks

But I think that there are some reasons why those fields can be empty, there's maybe a reason and we should handle that case.

Because for now, imagine if those checks fail and we return True, this is misleading because we did not actually set the secret uri while we are telling we did.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: OK fair, that's a limitation of python typer sadly.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: In class data, we have:
```python3
class Data(ABC):
    SECRET_FIELDS = []

    def __init__(self, ...):
        ...
        self._secret_fields = self.SECRET_FIELDS
```
So this field will be set correctly since ProviderData does not overload `SECRET_FIELDS`.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: OK :)

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: What I mean is that for any developer who knows the library, we're inverting the logic here between what is provided and what is requested and I'm pretty sure that's not something we want in terms of how we maintain the library lifecycle.
Otherwise, I think that we have to publish it as a new major because we're changing a lot of the logic and responsibilites. You said yourself that the changes are too big to stay consistent, for me that would be a reason for a new major, not a minor bump.

lib/charms/data_platform_libs/v0/data_interfaces.py
Gu1nness: The only additional test that I see is in `_delete_relation_data` and I think it's fair to assume that this check can always be done.
I have no strong opinion here however (I personally love decorators for this kind of cases because it does what you expect and it stands out in the code so you won't miss it).

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
Gu1nness: Ok :)

lib/charms/data_platform_libs/v0/data_interfaces.py
skourta: I have added this back as the [updated linter started to complain](https://github.com/canonical/data-platform-libs/actions/runs/13962134345/job/39085272116#step:4:11) about it. If you have a strong opinion on it then we can add an exception for it.

lib/charms/data_platform_libs/v0/data_interfaces.py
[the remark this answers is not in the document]
skourta: I reverted the initialisation logic to have `SECRET_FIELDS` represent the remote secret fields instead.

Gu1nness (approved):

reneradoi (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -3778,3 +3835,324 @@ def __init__(
...
+
+        return self.relation.data[self.relation.app].get("prefix")
+
+    @property
+    def mtls_chain(self) -> Optional[str]:
+        """Returns TLS chain of the client."""
+        if not self.relation.app:
+            return None
+
+        if self.secrets_enabled:
```
Mehdi-Bendriss, line 3860: This should always be `True` for Databases with mTLS support, no?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -3778,3 +3835,324 @@ def __init__(
...
+            secret_field = f"{PROV_SECRET_PREFIX}{SECRET_GROUPS.MTLS}"
+            if secret_uri := self.relation.data[self.app].get(secret_field):
+                secret = self.framework.model.get_secret(id=secret_uri)
+                content = secret.get_content(refresh=True)
+                if content:
+                    return content.get("mtls-chain")
+
+        return self.relation.data[self.relation.app].get("mtls-chain")
```
Mehdi-Bendriss, line 3868: Related to the above comment, we can have the mTLS flow purely secrets based.

Mehdi-Bendriss (commented):
Thanks Smail - I only reviewed the mTLS and ETCD parts of this PR, they looks good. 
I do not see a reason to add support for relation-based-secrets

Mehdi-Bendriss (approved):

Gu1nness (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1000,38 +1009,250 @@ def secret_label_map(self):
...
 
+    @property
+    def secret_fields(self) -> Optional[List[str]]:
+        """Local access to secrets field, in case they are being used."""
+        if self.secrets_enabled:
+            return self._secret_fields
+
+    @property
+    def remote_secret_fields(self) -> Optional[List[str]]:
+        """Local access to secrets field, in case they are being used."""
```
delgod, line 1020:
```suggestion
        """Remote access to secrets field, in case they are being used."""
```

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -978,6 +983,10 @@ def __init__(
         self.component = self.local_app if self.SCOPE == Scope.APP else self.local_unit
         self.secrets = SecretCache(self._model, self.component)
         self.data_component = None
+        self._secret_fields = []
+        self._remote_secret_fields = list(self.SECRET_FIELDS)
+        # TO BE REPLACED BY PROVIDER AND REQUIRER
+        self._my_secret_groups = []
```
delgod, line 989: I don't see how the data from this variable is used in the code. Maybe we can delete it from everywhere.
But separate PR, please

delgod (approved):

PR #211, opened 2025-03-31
[MISC] Empty fetch_relation_data on follower units

dragomirp:
`fetch_relation_data` will not return any values on non-leader units due to an overprotective check.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -331,7 +331,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 45
+LIBPATCH = 46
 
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1052,15 +1048,6 @@ def _legacy_apply_on_delete(self, fields: List[str]) -> None:
 
     # Internal helper methods
 
-    @staticmethod
-    def _is_relation_active(relation: Relation):
-        """Whether the relation is active based on contained data."""
-        try:
-            _ = repr(relation.data)
```
dragomirp, line 1059: The local app data is only accessible to the leader, on followers trying to access it will cause an exception and returning False here. In principle, we can do the check while trying to filter out the local app data, but I don't think we need this check at all.

marceloneppel (approved):
Thanks for the comment. It makes sense to remove that check.

Gu1nness (approved):

delgod (approved):

PR #222, opened 2025-06-02
[DPE-7365] feat: add mtls-cert to kafka_client

imanenami:
This PR adds `mtls-cert` field to `kafka_client` relations, compliant with DA150. This change is backwards-compatible since the field is optional and it also doesn't break clients which rely on positional args.

### DIFF
--- a/tests/v0/integration/kafka-charm/src/charm.py
+++ b/tests/v0/integration/kafka-charm/src/charm.py
@@ -9,16 +9,24 @@ of the libraries in this repository.
 """
 
 import logging
-from typing import Dict, Optional
+import os
+from typing import Dict, Optional, cast
 
-from ops.charm import ActionEvent, CharmBase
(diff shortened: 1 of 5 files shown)


tests/v0/integration/kafka-charm/src/charm.py
```diff
@@ -9,17 +9,25 @@
...
+from charms.data_platform_libs.v0.data_interfaces import (
+    LIBPATCH as DATA_INTERFACES_VERSION,
+)
 from charms.data_platform_libs.v0.data_interfaces import (
     KafkaProvides,
     TopicRequestedEvent,
 )
 
+if DATA_INTERFACES_VERSION > 46:
+    from charms.data_platform_libs.v0.data_interfaces import KafkaClientMtlsCertUpdatedEvent
```
dragomirp, line 28: Do we need the conditional import here?

PR #227, opened 2025-06-18
[DPE-7386] Role permissions in data-interfaces v0

sinclert-canonical:
This PR supports the specification of role permissions in the `data_interfaces` v0 charm library.

### Approach

The proposed approach defines a new custom event (`entity_permissions_changed`) in order to completely decouple the existing behavior when the interface is used to update some other field, from when the interface is used to tweak a custom role permissions.

### References

- Role permissions in data-interfaces v0 [spec](https://docs.google.com/document/d/1DfHPScwb83lvFIrwmzfTOJ5pq4rgV5RLpiAVSi0gcc8/edit?tab=t.0).

---

Depends on https://github.com/canonical/data-platform-libs/pull/218

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -2059,24 +2059,16 @@ class DataPeerData(RequirerData, ProviderData):
         self,
         model,
         relation_name: str,
-        extra_user_roles: Optional[str] = None,
         additional_secret_fields: Optional[List[str]] = [],
         additional_secret_group_mapping: Dict[str, str] = {},
         secret_field_name: Optional[str] = None,
         deleted_label: Optional[str] = None,
-        extra_group_roles: Optional[str] = None,
-        entity_type: Optional[str] = None,
-        entity_permissions: Optional[str] = None,
     ):
         RequirerData.__init__(
             self,
-            model,
-            relation_name,
-            extra_user_roles,
-            additional_secret_fields,
-            extra_group_roles,
-            entity_type,
-            entity_permissions,
+            model=model,
+            relation_name=relation_name,
+            additional_secret_fields=additional_secret_fields,
         )
         self.secret_field_name = secret_field_name if secret_field_name else self.SECRET_FIELD_NAME
         self.deleted_label = deleted_label
@@ -2593,28 +2585,20 @@ class DataPeer(DataPeerData, DataPeerEventHandlers):
         self,
         charm,
         relation_name: str,
-        extra_user_roles: Optional[str] = None,
         additional_secret_fields: Optional[List[str]] = [],
         additional_secret_group_mapping: Dict[str, str] = {},
         secret_field_name: Optional[str] = None,
         deleted_label: Optional[str] = None,
         unique_key: str = "",
-        extra_group_roles: Optional[str] = None,
-        entity_type: Optional[str] = None,
-        entity_permissions: Optional[str] = None,
     ):
         DataPeerData.__init__(
             self,
             charm.model,
             relation_name,
-            extra_user_roles,
             additional_secret_fields,
             additional_secret_group_mapping,
             secret_field_name,
             deleted_label,
-            extra_group_roles,
-            entity_type,
-            entity_permissions,
         )
         DataPeerEventHandlers.__init__(self, charm, self, unique_key)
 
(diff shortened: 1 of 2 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2047,6 +2065,7 @@ def __init__(
         deleted_label: Optional[str] = None,
         extra_group_roles: Optional[str] = None,
         entity_type: Optional[str] = None,
+        entity_permissions: Optional[str] = None,
```
imanenami, line 2068: **issue**: Like the comment I had on previous PR, I don't think we should add these new fields to peer and unit Data objects. they're not used there.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2620,6 +2642,7 @@ def __init__(
         unique_key: str = "",
         extra_group_roles: Optional[str] = None,
         entity_type: Optional[str] = None,
+        entity_permissions: Optional[str] = None,
```
imanenami, line 2645: **issue:** Similar comment as above

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1872,6 +1876,20 @@ def _validate_entity_type(self) -> None:
...
+
+        accepted_keys = {"resource_name", "resource_type", "privileges"}
+
+        try:
+            permissions = json.loads(self.entity_permissions)
+            for permission in permissions:
+                assert permission.keys() == accepted_keys
+        except json.decoder.JSONDecodeError:
+            raise ValueError("Invalid entity permissions format. It must be JSON format")
+
```
imanenami, line 1892: **question:** what's the expectation from developers, should they extend this and specific validation checks based on the data product? if so (and I think it should be), this method should not be private.

imanenami (commented):
Thanks, solid work. Please check the comments as I have some issues/questions regarding this extension.

lib/charms/data_platform_libs/v0/data_interfaces.py
sinclert-canonical: The only centralized place to put the `entity-type` (previous PR) and `entity-permissions` (current PR) validations  was the `RequirerData` class (see [code](https://github.com/canonical/data-platform-libs/blob/7eaf77ea8c2e01c4c91286d1fa2fd609fac6a9ad/lib/charms/data_platform_libs/v0/data_interfaces.py#L1837-L1838)). Such class does **not only** get initialized within _DatabaseRequires_, _KafkaRequires_, == and _OpensearchRequires_ classes, but also by the `DataPeerData` one (see [code](https://github.com/canonical/data-platform-libs/blob/7eaf77ea8c2e01c4c91286d1fa2fd609fac6a9ad/lib/charms/data_platform_libs/v0/data_interfaces.py#L2050-L2079)).

I did not wanted to tweak the class inheritance structure, so I needed to also pass the new arguments to the `DataPeerData` class init method, to keep the validation logic working. As you can see, I am not saving them as instance variables, I am just passing them down to the parent class so it can validate.

lib/charms/data_platform_libs/v0/data_interfaces.py
sinclert-canonical: Answered in [this comment](https://github.com/canonical/data-platform-libs/pull/227#discussion_r2215664709).

Let's continue the conversation there, in case you think we should change the implementation.

lib/charms/data_platform_libs/v0/data_interfaces.py
sinclert-canonical: Apologies for the misunderstanding, but no, **that is not the expectation**.

These fields were explicitly approved as part of the [spec](https://docs.google.com/document/d/1DfHPScwb83lvFIrwmzfTOJ5pq4rgV5RLpiAVSi0gcc8/edit?tab=t.0) that triggered the creation of this PR. They are supposed to be used, within the JSON passed as the `entity-permissions` argument, to specify a list of privileges for Juju cluster admins to configure. The design needed to be as fine-grained as possible to encode [this permissions file](https://git.launchpad.net/launchpad/tree/database/schema/security.cfg) from LaunchPad team.

So, this list of fields is not expected to be extended, unless another spec superseeds the current one.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1872,6 +1876,20 @@ def _validate_entity_type(self) -> None:
...
+        """Validates whether the provided entity permissions follow the right JSON format."""
+        if not self.entity_permissions:
+            return
+
+        accepted_keys = {"resource_name", "resource_type", "privileges"}
+
+        try:
+            permissions = json.loads(self.entity_permissions)
+            for permission in permissions:
+                assert permission.keys() == accepted_keys
```
dragomirp, line 1889: We should most probably also raise a ValueError here for consistency's sake.

sinclert-canonical: Fair enough, otherwise an `AssertionError` would be raised.

Changed in commit https://github.com/canonical/data-platform-libs/pull/227/commits/6b7638c95ced5adcd515434636fe52035bf8740d.

dragomirp (approved):

lib/charms/data_platform_libs/v0/data_interfaces.py
imanenami: Thanks for the explanation, but I think you could pass them (as `None`) to the super class without defining them on the peer/unit data object constructors. Having an `entity-type` or `entity-permission` on `DataPeerData` for example, makes me (or any ignorant dev) wonder what these option would do.

sinclert-canonical: You are right. I originally added these new fields _everywhere_ the `extra-user-roles` was present. Now, thanks to you, I realized such field should have never been added to classes `DataPeer`, `DataPeerUnit`...

I pruned them all in commit https://github.com/canonical/data-platform-libs/pull/227/commits/a9bd64bfd75354d490542480d122bf0f4fece92a.

PR #241, opened 2025-09-05
feat: Data Interfaces V1

Gu1nness:
Implementation of Data Interfaces V1:
 * Provides repositories, interfaces, helper methods
 * Secret encryption/decryption based on typing of fields in models
 * Missing a bit of documentation (will be added in the coming days)
 * Unit + Integration tests
 * Examples in the integration tests + documentation
 * Supports data interfaces (basic) + permissions + entity

This V1 supports multiple requests from the requirer, and supports both bulk events (1 event for all the requests at the same time) and single events (one event per request), according to your usecase.

It supports credentials updates from the provider (used by some charms), it supports MTLS from the requirer, and heavily relies on type systems and pydantic!

Example of use
```python3
from charms.data_platform_libs.v1.data_interfaces import (
    RequirerCommonModel,
    RequirerDataContractV1,
    ResourceCreatedEvent,
    ResourceEntityCreatedEvent,
    ResourceProviderModel,
    ResourceRequirerEventHandler,
)

class ApplicationCharm(CharmBase):
    # Application charm that connects to database charms.

    def __init__(self, *args):
        super().__init__(*args)

        requests = [
            RequirerCommonModel(
                resource="clientdb",
            ),
            RequirerCommonModel(
                resource="clientbis",
            ),
        ]
        # Define the cluster aliases and one handler for each cluster database created event.
        self.database = ResourceRequirerEventHandler(
            self,
            relation_name="database"
            relations_aliases = ["cluster1", "cluster2"],
            requests=
        )
         self.framework.observe(
             self.database.on.cluster1_resource_created, self._on_cluster1_resource_created
         )
         self.framework.observe(
             self.database.on.cluster2_resource_created, self._on_cluster2_resource_created
         )

     def _on_cluster1_resource_created(self, event: ResourceCreatedEvent) -> None:
         # Handle the created database on the cluster named cluster1

         # Create configuration file for app
         config_file = self._render_app_config_file(
             event.response.username,
             event.response.password,
             event.response.endpoints,
         )
         ...

     def _on_cluster2_resource_created(self, event: ResourceCreatedEvent) -> None:
         # Handle the created database on the cluster named cluster2

         # Create configuration file for app
         config_file = self._render_app_config_file(
             event.response.username,
             event.response.password,
             event.response.endpoints,
         )
         ...
```

### DIFF
--- a/lib/charms/data_platform_libs/v1/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v1/data_interfaces.py
@@ -2544,7 +2544,9 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
 
         model = self.interface.build_model(relation_id, DataContractV1[responses[0].__class__])
 
-        response_map: dict[str, ResourceProviderModel] = {response.request_id: response for response in responses if response.request_id}
+        response_map: dict[str, ResourceProviderModel] = {
+            response.request_id: response for response in responses if response.request_id
+        }
 
         # Update all the already existing keys
         for index, _response in enumerate(model.requests):
@@ -2562,9 +2564,7 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
 
     def requests(self, relation: Relation) -> Sequence[RequirerCommonModel]:
         """Returns the list of requests that we got."""
-        repository = OpsRelationRepository(
-            self.model, relation, component=relation.app
-        )
+        repository = OpsRelationRepository(self.model, relation, component=relation.app)
 
         # Don't do anything until we get some data
         if not repository.get_data():
@@ -2579,7 +2579,9 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
             request_model = build_model(repository, RequirerDataContractV1[self.request_model])
             return request_model.requests
 
-    def responses(self, relation: Relation, model: type[ResourceProviderModel]) -> list[ResourceProviderModel]:
+    def responses(
+        self, relation: Relation, model: type[ResourceProviderModel]
+    ) -> list[ResourceProviderModel]:
         """Returns the list of responses that we currently have."""
         repository = self.interface.repository(relation.id, component=relation.app)
 
@@ -2589,9 +2591,6 @@ class ResourceProviderEventHandler(EventHandlers, Generic[TRequirerCommonModel])
             return [self.interface.build_model(relation.id, DataContractV0)]
 
         return self.interface.build_model(relation.id, DataContractV1[model]).requests
-
-
-
 
 
 class ResourceRequirerEventHandler(EventHandlers, Generic[TResourceProviderModel]):


lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+
+        username = event.response.username
+        password = event.response.password
+        ...
+
+    def _on_resource_entity_created(self, event: ResourceCreatedEvent) -> None:
+        # Event triggered when a new entity is created.
+        ...
+
+Compared to V1, this library makes heavy use of pydantic models, and allows for
```
sinclert-canonical, line 95: I think there is a typo here:

```diff
- Compared to V1, this library makes heavy use of pydantic models, and allows for
+ Compared to V0, this library makes heavy use of pydantic models, and allows for
```

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+        ...
+```
+
+### Provider Charm
+
+Following an example of using the ResourceRequestedEvent, in the context of the
+database charm code:
+
+```python
+from charms.data_platform_libs.v0.data_interfaces import DatabaseProvides
```
sinclert-canonical, line 197: I think there is a typo here:

```diff
- from charms.data_platform_libs.v0.data_interfaces import DatabaseProvides
+ from charms.data_platform_libs.v1.data_interfaces import DatabaseProvides
```

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+def ensure_leader_for_app(f):
+    """Decorator to ensure that only leader can perform given operation."""
+
+    def wrapper(self, *args, **kwargs):
+        if self.component == self._local_app and not self._local_unit.is_leader():
+            logger.error(f"This operation ({f.__name__}) can only be performed by the leader unit")
+            return
+        return f(self, *args, **kwargs)
+
+    return wrapper
```
sinclert-canonical, line 376: Lovely!

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+
+    @abstractmethod
+    def delete_fields(self, *fields: str) -> None:
+        """Deletes all the provided fields."""
+        ...
+
+    @abstractmethod
+    def delete_secret_field(self, field: str, secret_group: SecretGroup) -> None:
+        """Delete a field stored in a secret group."""
+        ...
```
sinclert-canonical, line 1110: Suggestion: what about inverting `secret_group` and `field` arguments order?

It seems this method is closer to the ones below it (leveraging both `secret_group` and `field` args), than the ones above it (only leveraging `field` arg).

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+        """Generate unique group_mappings for secrets within a relation context."""
+        if short_uuid:
+            return f"{relation.name}.{relation.id}.{short_uuid}.{secret_group}.secret"
+        return f"{relation.name}.{relation.id}.{secret_group}.secret"
+
+    def secret_field(self, secret_group: SecretGroup, field: str | None = None) -> str:
+        """Generates the field name to store in the peer relation."""
+        return f"{self.SECRET_FIELD_NAME}-{secret_group}"
+
+    get_data = ensure_leader_for_app(OpsRepository.get_data)
```
sinclert-canonical, line 1452: Clarification: are we overriding the `get_data` method, by wrapping it with a decorator?

If so, I think it would be more clear if we just override the method in a more traditional manner. IMHO the current proposal can be easily overlooked.

```python
@override
@ensure_leader_for_app
def get_data(...)
  super().get_data(...)
```

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+
+        if relation.name != self.relation_name:
+            logging.info("Secret changed on wrong relation.")
+            return
+
+        remote_unit = None
+        for unit in relation.units:
+            if unit.app != self.charm.app:
+                remote_unit = unit
+                break
```
sinclert-canonical, line 2279: Suggestion: extract these lines to a helper function within the `EventHandlers` class

This is the 1st occurrence of this _search snippet_. I think we could extract this logic into a common place within the `EventHandlers` class, class that both snippet methods inherit from.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
```
sinclert-canonical, line 2623: Suggestion: extract these lines to a helper function within the `EventHandlers` class

This is the 2nd occurrence of this _search snippet_. I think we could extract this logic into a common place within the `EventHandlers` class, class that both snippet methods inherit from.

tox.ini
```diff
@@ -136,6 +158,20 @@ deps =
 commands =
     pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/v0/integration/test_kafka_charm.py
 
+[testenv:integration-kafka-v1]
+description = Run Kafka integration tests
+deps =
+    psycopg2-binary
```
sinclert-canonical, line 164: Clarification: do we need `psycopg2-binary` for these integration tests?

sinclert-canonical (commented):
🧡 Extraordinary effort.

I left some suggestions in addition to the following general questions:
1. Is Python 3.10+ a requirement for this charm lib? Not that I am against, it was about time, but I know certain charms are stuck on Python 3.8 (PGBouncer VM is one, as an example).
2. Where do you think it would be a good place to rescue [this](https://github.com/canonical/data-platform-libs/blob/3c4f5d4df8a66bb09efbe2fd5ee8bbf3bea17d3a/lib/charms/data_platform_libs/v0/data_interfaces.py#L1998-L2011) validation function from the V0 library? I noticed it is not present on the current proposal.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+class OpsPeerRepository(OpsRepository):
+    """Implementation of the Ops Repository for peer relations."""
+
+    SECRET_FIELD_NAME = "internal_secret"
+
+    IGNORES_GROUPS = [
+        SecretGroup("user"),
+        SecretGroup("entity"),
+        SecretGroup("mtls"),
+        SecretGroup("tls"),
```
reneradoi, line 1464: Do we need to consider `extra` here?

reneradoi (commented):
Hey @Gu1nness this is impressive work! Very well designed, thoughtful of details and I really like the rich set of examples and comments. I don't have comments on implementation details, it all makes a lot of sense to me as far as I can follow. Some questions for understanding and one thing to clarify (is `prefix` still supported?), other than that it is well rounded already.

lib/charms/data_platform_libs/v1/data_interfaces.py
Gu1nness: Excellent suggestion !

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+                    )
+                    return cursor.fetchone() is not None
+        except psycopg2.Error as e:
+            logger.exception(
+                f"failed to check whether {plugin} plugin is enabled in the database: %s",
+                str(e),
+            )
+            return False
+
+    def is_postgresql_plugin_enabled(self, plugin: str, relation_index: int = 0) -> bool:
```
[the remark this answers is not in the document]
Gu1nness: It pains me, but I asked and apparently it has to stay indeed.
Trust me, I'd love to see that removed…

sinclert-canonical (approved):

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+
+
+class Scope(Enum):
+    """Peer relations scope."""
+
+    APP = "app"
+    UNIT = "unit"
+
+
+class CachedSecret:
```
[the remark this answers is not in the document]
reneradoi: Ok thank you. If I understand correctly, removing the cache at a later point should not break backwards compatibility, so we can keep it for now and do this in a future PR, if it makes sense to others too.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2750 @@
...
+
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_created,
+            self._on_relation_created_event,
+        )
+
+        self.framework.observe(
+            charm.on.secret_changed,
+            self._on_secret_changed_event,
+        )
```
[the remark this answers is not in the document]
reneradoi: The comment was rather about cleaning up old revisions, not the secrets themselves. There has been a bit of noise about this topic lately, as this is obviously also not done in V0 and led to production impact on Postgres.

[Here's](https://documentation.ubuntu.com/juju/3.6/reference/secret/#secret-lifecycle) what's in the docs: `Charms that create secrets should always handle the secret-remove event. That is because secret revisions, even if obsolete, remain until removed by the charm; if a charm does not remove them, they accumulate indefinitely.`

Looks like this is not really covered yet, anywhere. Would make sense from my point of view to add this to the lib itself.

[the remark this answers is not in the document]
Gu1nness: Oh right.
Yeah I'll add it

lib/charms/data_platform_libs/v1/data_interfaces.py
reneradoi: I don't have a strong opinion, we can leave it as is. Thank you for the clarification.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2818 @@
...
+            return
+        relation = self._relation_from_secret_label(event.secret.label)
+
+        if not relation:
+            logging.info(
+                f"Received secret {event.secret.label} but couldn't parse, seems irrelevant"
+            )
+            return
+
+        if relation.app != self.charm.app:
```
skourta: We missed updating this yesterday

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2818 @@
...
+                    if value:
+                        secret = repository.add_secret(
+                            aliased_field, actual_value, secret_group, short_uuid
+                        )
+                        if not secret or not secret.meta:
+                            raise SecretError("No secret to send back")
+                        setattr(self, secret_field, secret.meta.id)
+                    continue
+
+                if secret and secret.meta:
```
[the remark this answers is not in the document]
skourta: I think we should still update this regardless of `set_response`.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -0,0 +1,2832 @@
...
+        return f"{'.'.join(members)}"
+
+    def secret_field(self, secret_group: SecretGroup, field: str | None = None) -> str:
+        """Generates the field name to store in the peer relation."""
+        if not field:
+            raise ValueError("Must have a field.")
+        return f"{field}@{secret_group}"
+
+
+@final
```
zmraul: Is `final` on these classes needed? I think a valid use case is the one we are mentioning on other comments about extending the repositories with joint access to data/secrets and having a dict-like `update` method.

Even if these use cases are baked into the lib, we could still find extensions to how the data is modeled at some point in the future.

Gu1nness: I think they can be removed.

marceloneppel:
> > Thanks for the amazing job done here, @Gu1nness!
> > I started to test it against the [PostgreSQL 16 K8s](https://github.com/canonical/postgresql-k8s-operator/pull/1129) on CI and locally (it's still a work in progress, but already given me more confidence in the upgrade of the library).
> > I have two questions:
> > ```
> > * Is the V1 library supposed to also support Juju 2.9 charms or only Juju 3.x charms?
> > ```
> 
> This is only for Juju 3 charms
> 
> > ```
> > * Is this V1 library compatible with older versions of the V0 library? I mean, there is some code like the one from [`3c4f5d4`/lib/charms/data_platform_libs/v0/data_interfaces.py#L2507C9-L2507C64](https://github.com/canonical/data-platform-libs/blob/3c4f5d4df8a66bb09efbe2fd5ee8bbf3bea17d3a/lib/charms/data_platform_libs/v0/data_interfaces.py#L2507C9-L2507C64), which migrates from an old internal secret to the structure we have today. Is it supported?
> > ```
> 
> There's is no backward compatibility with pre-secret charms, so no compatibility with the older versions of the library, this was explicitely requested by Mykola. I think in your case (postgresql), it means you can use it for Postgresql 16 but probably not for Postgresql 14 because of the juju 2.9 compatibility.

Thanks for the clarification.

skourta:
> +1 here from Cassandra folks! 🥳
> 
> We have integrated the library successfully on [cassandra](https://github.com/canonical/cassandra-operator/pull/21#pullrequestreview-3349849668). Is there any timeline when this PR is going to be merged or should we just go ahead and merge the Cassandra PR? I suppose we can always update the lib once it is consolidated

@deusebio any reason why you did not migrate the peer relation data handling to v1?

deusebio:
> any reason why you did not migrate the peer relation data handling to v1?

I'd say to keep the PR confined and capacity. My assumption (also looking at the spec DA175) is that data-interfaces were mostly concerned (and critical) for client<>server relations. Of course we could also apply them to peer, but that's nice to have for us right now. We have a few peer relations, so this may not be a trivial change. The current task we need to deliver before end of cycle are implementing client relations (the PR attached) and implement HA tests. There is not much time left before the end of the cycle and we would like to wrap those up. 

TLDR I believe using a more structured approach for peer can also be useful, but probably to be addressed in the future

reneradoi (approved):
Thank you @Gu1nness for all your effort and your work on this PR! I approve, no blockers from my side.

zmraul:
> @zmraul Looks better ?

Yes, thank you! This is a nice improvement :)

imanenami (approved):
This is already fabulous for the first release of V1, and we have integrated it successfully with the main Kafka charm, so I'm happy with it. Great effort @Gu1nness!

PR #242, opened 2025-09-10
[DPE-8033] feat: add status propagation to provider interface

imanenami:


sinclert-canonical (approved):
Great work! Just left a couple of comments.

Gu1nness (approved):
Love the implementation, love the simplicity!
Could we use the statuses on a more complex relation so that we see how that would work in case of a real case scenario?
Aside from that, it's really a clean implementation!

dragomirp (approved):

PR #257, opened 2026-02-06
fix: Don't listen to secret changed of other rels

Gu1nness:
## Issue

The secret changed event handlers are a bit too eager to handle secret changed events.
They should beware to handle only the events for the relations they are monitoring.

##

This commit enforces that on Data interfaces v0.
It works by checking the relation name computed from the secret with the relation name of the relation that we are monitoring.
If they are different, exit early and let the correct handler handle that event.

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -4276,7 +4276,7 @@ class KafkaProviderEventHandlers(ProviderEventHandlers):
             logging.info("Secret changed event ignored for Secret Owner")
 
         if relation.name != self.relation_data.relation_name:
-            logger.info(
+            logger.debug(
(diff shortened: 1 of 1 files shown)


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -4275,6 +4275,14 @@ def _on_secret_changed_event(self, event: SecretChangedEvent):
         if relation.app == self.charm.app:
             logging.info("Secret changed event ignored for Secret Owner")
 
+        if relation.name != self.relation_data.relation_name:
+            logger.info(
```
dragomirp: Maybe debug level would be better here, since it's expected behaviour.

dragomirp (approved):

marceloneppel (approved):
LGTM! Thanks!

PR #239, opened 2025-08-12
[DPE-7379] Requested entity name

dragomirp:
Implement a way to request entity name by passing a secret.

Draft implementation PRs:
* [Postgresql K8s](https://github.com/canonical/postgresql-k8s-operator/pull/1060)
* [Data integrator](https://github.com/canonical/data-integrator/pull/171)

PR #258, opened 2026-02-11
[DPE-8863] Custom username and prefixes for v1

dragomirp:
Port of https://github.com/canonical/data-platform-libs/pull/239 and https://github.com/canonical/data-platform-libs/pull/245 for DPLv1

* Add custom username support
* Add prefix resources support


lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -833,9 +836,16 @@ def extract_secrets(self, info: ValidationInfo):
...
+                try:
+                    secret = repository.get_secret(
+                        secret_group, secret_uri=secret_uri, short_uuid=short_uuid
+                    )
+                except SecretNotFoundError:
+                    # v0 deletes the requested entity secret
+                    if secret_group == "requested-entity":
+                        logger.debug("Missing requested entity secret")
+                        continue
+                    raise
```
dragomirp, line 848: v0 will delete the helper secret once the relation was established.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -1002,6 +1012,13 @@ class RequirerCommonModel(CommonModel):
     entity_permissions: list[EntityPermissionModel] | None = Field(default=None)
     secret_mtls: SecretString | None = Field(default=None)
     mtls_cert: MtlsSecretStr = Field(default=None)
+    secret_requested_entity: SecretString | None = Field(
+        default=None,
+        validation_alias=AliasChoices("requested-entity-secret", "secret-requested-entity"),
```
dragomirp, line 1017: Should match both the old and new secret field.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -1047,6 +1067,7 @@ class ResourceProviderModel(ProviderCommonModel):
     entity_name: EntitySecretStr = Field(default=None)
     entity_password: EntitySecretStr = Field(default=None)
     version: str | None = Field(default=None)
+    prefix_resources: str | None = Field(default=None)
```
dragomirp, line 1070: I wanted to use [PlainSerializer](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.PlainSerializer) to sort the CSV list, but it seems that the `field_info.annotation in OptionalSecrets` traps other annotations as well. Leaving it be for now.

tests/v1/integration/database-charm/src/charm.py
```diff
@@ -208,7 +208,7 @@ def _on_database_pebble_ready(self, event: WorkloadEvent) -> None:
                     "command": "/usr/local/bin/docker-entrypoint.sh postgres",
                     "startup": "enabled",
                     "environment": {
-                        "PGDATA": "/var/lib/postgresql/data/pgdata",
+                        "PGDATA": "/var/lib/postgresql/data/pgdata/data",
```
dragomirp, line 211: For Canonical k8s compatibility (non-empty volume is provided).

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -2632,6 +2660,11 @@ def set_response(self, relation_id: int, response: ResourceProviderModel):
             self.interface.write_model(
                 relation_id, response, context={"version": "v0"}
             )  # {"database": "database-name", "secret-user": "uri", ...}
+            # Set expected prefix field if present
+            if response.prefix_resources:
+                self.interface.repository(relation_id).write_field(
+                    "prefix-databases", response.prefix_resources
+                )
```
dragomirp, line 2667: Is there a better way to inject this field?

lib/charms/data_platform_libs/v1/data_interfaces.py
Gu1nness: Interesting.
Maybe we can add just a field serializer for this one in this class ? Or that prevents it?

lib/charms/data_platform_libs/v1/data_interfaces.py
Gu1nness: Oh, this is because the field is renamed ?
Then probably we can do like we do for the other resources: [here](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v1/data_interfaces.py#L1061)

Gu1nness (commented):
All seems good.
Let's investigate on the CSV ordering.
For the `prefix-database` vs `prefix-resource` let's try like we do for the resource field.
Same pattern, a dedicated field that we don't serialize just to store the original field name.

lib/charms/data_platform_libs/v1/data_interfaces.py
```diff
@@ -1002,6 +1012,13 @@ class RequirerCommonModel(CommonModel):
     entity_permissions: list[EntityPermissionModel] | None = Field(default=None)
     secret_mtls: SecretString | None = Field(default=None)
     mtls_cert: MtlsSecretStr = Field(default=None)
+    secret_requested_entity: SecretString | None = Field(
+        default=None,
+        validation_alias=AliasChoices("requested-entity-secret", "secret-requested-entity"),
+    )
+    entity_name: RequestedEntitySecretStr = Field(default=None)
+    entity_password: RequestedEntitySecretStr = Field(default=None)
```
marceloneppel: Hi, @dragomirp! I noticed a potential backward-compatibility issue here: the v0 library stores the requested entity password under the key `"password"` in the Juju secret ([v0 line 3855](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py#L3855)), but `extract_secrets` will look for `"entity-password"` (derived from this field name via the alias generator). This means when a v0 requirer sends a custom password, the v1 provider will silently get `None` instead.

lib/charms/data_platform_libs/v1/data_interfaces.py
dragomirp: What I tried to do was: https://github.com/canonical/data-platform-libs/blob/4a7774a5bb23d63478c61b9271c123610721f09a/lib/charms/data_platform_libs/v1/data_interfaces.py#L1070

With the serialiser being: https://github.com/canonical/data-platform-libs/blob/4a7774a5bb23d63478c61b9271c123610721f09a/lib/charms/data_platform_libs/v1/data_interfaces.py#L1032-L1036

But then the field got turned into a secret. I guess that the `OptionalSecretStr` alias traps this, since they are both `str | None`, but didn't dig further, since autosorting is nice to have, but not strictly necessary.

lib/charms/data_platform_libs/v1/data_interfaces.py
dragomirp: I think the `original_field` is still written in the provider response with `write_field` here: https://github.com/canonical/data-platform-libs/blob/ba0faad1bf8d52409aff80f96b4163935264bd40/lib/charms/data_platform_libs/v1/data_interfaces.py#L2588-L2596

lib/charms/data_platform_libs/v1/data_interfaces.py
dragomirp: Updated in https://github.com/canonical/data-platform-libs/pull/258/changes/9e37b6352072efc6a2139ef68d2df07414910c77.

marceloneppel (approved):

PR #201, opened 2024-12-04
[MISC] Create data-models version 1

sinclert-canonical:
This PR creates version 1 of those libraries that would suffer breaking changes due to the support of Pydantic 2:
- `data_models`.
- ~`upgrade`~.

Given that this is the first PR that bumps the major version of some of the libraries, some code shuffling was required to give the repository a cohesive view of what is still in `v0`, and what is available both in `v0` and `v1`.

---

### Pydantic 2 specific changes:
Pydantic 2 breaking changes can be seen in commit https://github.com/canonical/data-platform-libs/pull/201/commits/4fbfe4f8877340dc0f06ce72fc596a106ad8e547.

<details>
  <summary>Code substitutions</summary>

  ```diff
  - from pydantic import validator
  + from pydantic import field_validator

  - from pydantic import root_validator
  + from pydantic import model_validator

  - field.outer_type_
  + field.annotation

  - obj.dict
  + obj.model_dump

  - obj.__fields__
  + obj.model_fields

  - type(self.dependency_model)(**json.loads(deps))
  + type(self.dependency_model).model_validate_json(deps)

  - json.dumps(self.dependency_model.dict())
  + self.dependency_model.model_dump_json()
  ```

</details>

---

### Code shuffling changes:

- Moved + adapted all previous integration tests into [tests/v0/integration](https://github.com/canonical/data-platform-libs/tree/sinclert/pydantic-2/tests/v0/integration).
- Moved + adapted all previous unit tests into [tests/v0/unit](https://github.com/canonical/data-platform-libs/tree/sinclert/pydantic-2/tests/v0/unit).
- Copied + adapted `data_models.py` ~and `upgrade.py`~ tests into [tests/v1/unit](https://github.com/canonical/data-platform-libs/tree/sinclert/pydantic-2/tests/v1/unit).
- Copied + adapted `data_models.py` ~and `upgrade.py`~ lib into [libs/charms/data_platform_libs/v1](https://github.com/canonical/data-platform-libs/tree/sinclert/pydantic-2/lib/charms/data_platform_libs/v1).
- Adapted Tox.ini environments to allow for `-v0` and `-v1` suffixes.
- Adapted CI workflow to use the new Tox environments.

PR #260, opened 2026-03-02
Add 'upgrade' library v1

Pepsiqqq:
Adds upgrade v1 which uses Pydantic 2.0+ instead of version <2.
The unit test remains same as v0 because changes between two are minimal.
Key Changes Made:

1. Simplified validate_constraint: Because we bound `VersionConstraint` directly to the str type, Pydantic will pass a string into `validate_constraint` function. All the `isinstance` checks  are removed.
2. Replaced `@root_validator` with `@model_validator(mode="after")`, because it runs after field validation, `self` is fully constructed.
3. Dropped `skip_on_failure=True`, Pydantic v2 automatically skips mode="after" model validators if any individual field validations fail
4. Updated requirements to include poetry-core
5. Replaced `self.dependency_model.__fields__.keys()` with `self.dependency_model.model_fields.keys()`
6. Replaced `self.dependency_model.dict()` with `self.dependency_model.model_dump_json()`
7. Fixed typos in database charm in integration tests(to fix linter error)
8. Implemented IntEnum for upgrade state

### DIFF
--- a/lib/charms/data_platform_libs/v1/upgrade.py
+++ b/lib/charms/data_platform_libs/v1/upgrade.py
@@ -1,0 +1,1091 @@
+# Copyright 2026 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""Library to manage in-place upgrades for charms running on VMs and K8s.
+
+This library contains handlers for `upgrade` relation events used to coordinate
+between units in an application during a `juju refresh`, as well as `Pydantic` models
+for instantiating, validating and comparing dependencies.
+
+An upgrade on VMs is initiated with the command `juju refresh`. Once executed, the following
+events are emitted to each unit at random:
+    - `upgrade-charm`
+    - `config-changed`
+    - `leader-settings-changed` - Non-leader only
+
+Charm authors can implement the classes defined in this library to streamline the process of
+coordinating which unit updates when, achieved through updating of unit-data `state` throughout.
+
+At a high-level, the upgrade steps are as follows:
+    - Run pre-checks on the cluster to confirm it is safe to upgrade
+    - Create stack of unit.ids, to serve as the upgrade order (generally workload leader is last)
+    - Start the upgrade by issuing a Juju CLI command
+    - The unit at the top of the stack gets permission to upgrade
+    - The unit handles the upgrade and restarts their service
+    - Repeat, until all units have restarted
+
+### Usage by charm authors
+
+#### `upgrade` relation
+
+Charm authors must implement an additional peer-relation.
+
+As this library uses relation data exchanged between units to coordinate, charm authors
+need to add a new relation interface. The relation name does not matter.
+
+`metadata.yaml`
+```yaml
+peers:
+  upgrade:
+    interface: upgrade
+```
+
+#### Dependencies JSON/Dict
+
+Charm authors must implement a dict object tracking current charm versions, requirements + upgradability.
+
+Many workload versions may be incompatible with older/newer versions. This same idea also can apply to
+charm or snap versions. Workloads with required related applications (e.g Kafka + ZooKeeper) also need to
+ensure their versions are compatible during an upgrade, to avoid cluster failure.
+
+As such, it is necessasry to freeze any dependencies within each published charm. An example of this could
+be creating a `DEPENDENCIES` dict within the charm code, with the following structure:
+
+`src/literals.py`
+```python
+DEPENDENCIES = {
+    "kafka_charm": {
+        "dependencies": {"zookeeper": ">50"},
+        "name": "kafka",
+        "upgrade_supported": ">90",
+        "version": "100",
+    },
+    "kafka_service": {
+        "dependencies": {"zookeeper": "^3"},
+        "name": "kafka",
+        "upgrade_supported": ">=0.8",
+        "version": "3.3.2",
+    },
+}
+```
+
+The first-level key names are arbitrary labels for tracking what those versions+dependencies are for.
+The `dependencies` second-level values are a key-value map of any required external applications,
+    and the versions this packaged charm can support.
+The `upgrade_suppported` second-level values are requirements from which an in-place upgrade can be
+    supported by the charm.
+The `version` second-level values correspond to the current version of this packaged charm.
+
+Any requirements comply with [`poetry`'s dependency specifications](https://python-poetry.org/docs/dependency-specification/#caret-requirements).
+
+### Dependency Model
+
+Charm authors must implement their own class inheriting from `DependencyModel`.
+
+Using a `Pydantic` model to instantiate the aforementioned `DEPENDENCIES` dict gives stronger type safety and additional
+layers of validation.
+
+Implementation just needs to ensure that the top-level key names from `DEPENDENCIES` are defined as attributed in the model.
+
+`src/upgrade.py`
+```python
+from pydantic import BaseModel
+
+class KafkaDependenciesModel(BaseModel):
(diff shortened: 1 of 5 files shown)


sinclert-canonical (dismissed):
👋🏻 Hey! Thanks for your contribution.

Something similar to these changes was proposed back in 2024, and got a negative response out of Data Platform leadership (see how the _upgrade library_ was listed and crossed out in [this PR](https://github.com/canonical/data-platform-libs/pull/201) description).

Back then, the Data Platform department made a **conscious and strategic decision** not to have a V1 upgrade library. IIRC, the reason was related to how all efforts regarding `juju refresh` behavior should be redirected towards "Refresh V3", the modern equivalent of this library (I know, naming is confusing).

Mehdi-Bendriss:
Hey @sinclert-canonical - good comment! 
For context, this was brought to leadership last week, and we agreed we would have a refresh v1 that is equivalent to refresh v0, with the only difference being Pydantic v2. 

This follows a similar approach to `data_models v1` where the only difference with the v0 is pydantic v2 support.

The reasoning is that some apps (non sophisticated) unlike databases, don't need the complexity of refresh v2, v3. And were already on refresh v0. Those apps now face a bit of a holdup in their progression, as most of their dependencies and libs (advanced statuses for ex) rely heavily on Pydantic v2.

sinclert-canonical:
> For context, this was brought to leadership last week, and we agreed we would have a refresh v1 that is equivalent to refresh v0, with the only difference being Pydantic v2.

Interesting outcome.

Then @Pepsiqqq please validate your Upgrade V1 proposal with [these changes](https://github.com/canonical/data-platform-libs/pull/201/changes/dc85e3a8347e24a1571b48b53dffaa5e717a149e#diff-2b354b36bb55fcf698717b5cae104b7f0a77f06b402ab6c207aee1865404ea00) proposed in the 2024 PR, just to be sure you are not missing anything.

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1094 @@
...
+            )
+
+    def _on_pre_upgrade_check_action(self, event: ActionEvent) -> None:
+        """Handler for `pre-upgrade-check-action` events."""
+        if not self.peer_relation:
+            event.fail(message="Could not find upgrade relation.")
+            return
+
+        if not self.charm.unit.is_leader():
+            event.fail(message="Action must be ran on the Juju leader.")
```
gatici:
```suggestion
            event.fail(message="Action must be run on the Juju leader.")
```

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    def _on_upgrade_finished(self, _) -> None:
+        """Handler for `upgrade-finished` events."""
+        if self.substrate == "vm" or not self.peer_relation:
+            return
+
+        # Emit the upgrade relation changed event in the leader to update the upgrade_stack.
+        if self.charm.unit.is_leader():
+            self.charm.on[self.relation_name].relation_changed.emit(
+                self.model.get_relation(self.relation_name)
```
gatici:
```suggestion
                self.charm.model.get_relation(self.relation_name)
```
DataUpgrade class does not have a self.model.

gatici (commented):
@Pepsiqqq  Nice work. A few minor things are added.

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1094 @@
...
+    def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
+        """Handler for `upgrade-granted` events.
+
+        Handlers of this event must meet the following:
+            - SHOULD check for related application deps from :class:`DataUpgrade.dependencies`
+                - MAY raise :class:`DependencyError` if dependency not met
+            - MUST update unit `state` after validating the success of the upgrade, calling one of:
+                - :class:`DataUpgrade.set_unit_failed` if the unit upgrade fails
+                - :class:`DataUpgrade.set_unit_completed` if the unit upgrade succeeds
+            - MUST call :class:`DataUpgarde.on_upgrade_changed` on exit so event not lost on leader
```
gatici:
```suggestion
            - MUST call :class:`DataUpgrade.on_upgrade_changed` on exit so event not lost on leader
```

tests/v1/unit/test_upgrade.py
```diff
@@ -0,0 +1,1103 @@
...
+        charm=harness.charm, dependency_model=GandalfModel(**GANDALF_DEPS), substrate="k8s"
+    )
+    harness.add_relation("upgrade", "gandalf")
+
+    assert not harness.charm.upgrade.stored_dependencies
+
+    harness.charm.upgrade.upgrade_stack = []
+    harness.charm.upgrade.peer_relation.data[harness.charm.unit].update({"state": "completed"})
+    harness.add_relation_unit(harness.charm.upgrade.peer_relation.id, "gandalf/1")
+    harness.set_leader()
```
gatici:
```suggestion
    harness.set_leader(True)
```

tests/v1/unit/test_upgrade.py
```diff
@@ -0,0 +1,1103 @@
...
+    assert harness.charm.upgrade.stored_dependencies == GandalfModel(**GANDALF_DEPS)
+
+
+def test_upgrade_changed_sets_idle_and_deps_if_some_completed_idle(harness):
+    harness.charm.upgrade = GandalfUpgrade(
+        charm=harness.charm, dependency_model=GandalfModel(**GANDALF_DEPS), substrate="k8s"
+    )
+    harness.add_relation("upgrade", "gandalf")
+
+    assert not harness.charm.upgrade.stored_dependencies == GandalfModel(**GANDALF_DEPS)
```
gatici:
```suggestion
    assert  harness.charm.upgrade.stored_dependencies != GandalfModel(**GANDALF_DEPS)
```

gatici: This is more readable.

tests/v1/unit/test_upgrade.py
```diff
@@ -0,0 +1,1103 @@
...
+
+    harness.charm.upgrade.upgrade_stack = []
+    harness.charm.upgrade.peer_relation.data[harness.charm.unit].update({"state": "completed"})
+    harness.add_relation_unit(harness.charm.upgrade.peer_relation.id, "gandalf/1")
+    harness.set_leader()
```
gatici:
```suggestion
    harness.set_leader(True)
```

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1094 @@
...
+
+        Raises:
+            :class:`VersionError` if upgrading to existing `version` is not supported
+        """
+        keys = self.dependency_model.model_fields.keys()
+
+        compatible = True
+        incompatibilities: List[Tuple[str, str, str, str]] = []
+        for key in keys:
+            old_dep: DependencyModel = getattr(self.stored_dependencies, key)
```
gatici: self.stored_dependencies can be None and getattr  with None will raise.

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1094 @@
...
+    upgrade_finished = EventSource(UpgradeFinishedEvent)
+
+
+# --- EVENT HANDLER ---
+
+
+class DataUpgrade(Object, ABC):
+    """Manages `upgrade` relation operations for in-place upgrades."""
+
+    STATES = ["recovery", "failed", "idle", "ready", "upgrading", "completed"]
```
gatici: Adding and ordering states are error prone. Can we use Enums with an explicit ordering such as IntEnum ?

lib/charms/data_platform_libs/v1/upgrade.py
```diff
@@ -0,0 +1,1145 @@
...
+            ),
+        )
+
+    def restart(self, event) -> None:
+        if not self.upgrade.state == "idle":
+            event.defer()
+            return None
+
+        self.restart_snap_service()
+```
```
akram09, line 260: I would suggest to add a text clarifying that this is porting the v0 version to use Pydantic to add more context.

akram09 (approved):
Seems good for me

Mehdi-Bendriss (approved):

gatici (approved):

Gu1nness (approved):
Everything looks good to me, clean use of pydantic v2!

PR #266, opened 2026-05-27
[DPE-10184] fix mtls certificate for cross-model-relations

reneradoi:
This PR fixes an issue when providing the mTLS certificate in a cross-model relation.

Due to https://github.com/juju/juju/issues/21248#issuecomment-3564906360, granting a Juju secret from the requirer side of a cross-model relations fails with `sharing consumer secrets across a cross model relation not supported`. The proposed workaround from [DA264](https://docs.google.com/document/d/1kVrGGBezI3nxmow_dwNbSLmQsz6092vhbbeha2wyNGI/edit?usp=sharing) is to place the `mtls-cert` into regular relation data instead of a secret, if the relation is a cross-model relation.

The scope of this PR is to fix this issue for `data-interfaces` v0. The required fixes for v1 will be provided in a separate PR to https://github.com/canonical/data-platform-charmlibs (see https://github.com/canonical/data-platform-charmlibs/pull/16).

**Changes:**
- add dependency to `cryptography` for using `Fernet`
- add list `CROSS_MODEL_RELATION_CONSUMER_SECRETS` (currently only includes `mtls-cert`
- ignore keys defined in `CROSS_MODEL_RELATION_CONSUMER_SECRETS` for requirer-side secrets in cross-model relations
- add handling for `relation-created` event in `ProviderEventHandlers` class: in cross-model relations, generate the encryption key, store it in a secret, grant the secret to the requirer and add the secret id to relation data
- add logic for cross-model relations in `_update_relation_data_without_secrets`: read the encryption key from the secret `encryption-secret` and encrypt keys defined in `CROSS_MODEL_RELATION_CONSUMER_SECRETS`; in case of failures, values will be left blank
- add logic for cross-model relations in `_fetch_relation_data_without_secrets`: read the encryption key from the secret `encryption-secret` and decrypt keys defined in `CROSS_MODEL_RELATION_CONSUMER_SECRETS`
- adjust the `relation-changed` events in `EtcdRequirerEventHandlers` and `KafkaRequirerEventHandlers` classes: resend a previous request if the encryption secret was added to relation data from provider side, to ensure data is sent correctly

**Testing:**
Functionality was tested manually and with integration test coverage on charmed-etcd (see [this draft PR](https://github.com/canonical/charmed-etcd-operator/pull/236), example CI run [here](https://github.com/canonical/charmed-etcd-operator/actions/runs/27139989628/job/80122999043?pr=236#step:11:25)). Integration test coverage for this repo can be added after changes where merged in [this etcd PR](https://github.com/canonical/charmed-etcd-operator/pull/236) (because a provider charm with mtls support is needed)
- add `cryptography` to deps for all test charms
- fix the broken `s3-app` test charm
- remove `check_logs` where possible (only if action results are ensured otherwise) to avoid test flakyness


lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1586,11 +1591,41 @@ def _fetch_relation_data_without_secrets(
...
-            return {
+            relation_data = {
                 k: relation.data[component][k] for k in fields if k in relation.data[component]
             }
         else:
-            return dict(relation.data[component])
+            relation_data = dict(relation.data[component])
+
+        try:
+            remote_model = relation.remote_model.uuid
```
Gu1nness: Nit: this could be named `remote_model_uuid` to match what it is.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1586,11 +1591,41 @@ def _fetch_relation_data_without_secrets(
...
+        except (FileNotFoundError, ModelError, RuntimeError):
+            # access to remote model added in Juju 3.6.2, fails with 2.9
```
Gu1nness, line 1616: What is the error raised in case of Juju 2.9 ?
Maybe we could have a log line for that specific case?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2389,6 +2469,38 @@ def _validate_entity_consistency(event: RelationEvent, diff: Diff) -> None:
...
+            return
+
+        if self.model.uuid == remote_model:
+            return
+
+        # in cross-model relations, generate an encryption key and share it with the requirer as a secret
+        event_data = {}
+        secret_label = f"{self.model.uuid}-{event.relation.id}-encryption-secret"
```
Gu1nness, line 2520: What is the need for the model uuid in the secret label ?

tests/v0/integration/application-charm-etcd-client/requirements.txt
Gu1nness, line 1: There's an inconsistency / mismatch between this file's cryptography version and the one above (`requirements/v0/requirements.txt`)
I think for consistency it's good to keep only one.

tests/v0/integration/s3-charm/metadata.yaml
```diff
@@ -7,9 +7,6 @@ summary: |
   Data platform libs s3 meant to be used
   only for testing of the libs in this repository.
 
-series:
-  - focal
```
Gu1nness, line 11: Isn't this for juju 2.9 purposes?

tests/v0/integration/test_charm.py
Gu1nness, line 1: What's the reason of all this cleanup ?
Where those tests testing nothing?

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Updated

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: I've added a log line.

According to [ops](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Relation.remote_model), this raises a `ModelError`. But I've also seen instances of `FileNotFoundError` and `RuntimeError` in integration test runs, that's why I added those.

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Mostly consistency with another instance of creating a "helper" secret [here](https://github.com/canonical/data-platform-libs/blob/3f4aa1ad68b91b1011b4edbb2f0f0397a8b0b188/lib/charms/data_platform_libs/v0/data_interfaces.py#L3968-L3970).

tests/v0/integration/application-charm-etcd-client/requirements.txt
reneradoi: Thanks, I've cleaned that. It is a leftover from starting with `48.0.0` until I noticed that it is not available on all earlier platforms, then downgraded.

tests/v0/integration/s3-charm/metadata.yaml
reneradoi: Well, interesting one: I think this is a mistake (or leftover from the past). Even when deployed on `jammy` juju container, the container image was always `focal`, which resulted in issues when using the default Python of the container. It took quite some time to figure out why integration tests started to fail since the last commit and this repo and my PR.

Building for different bases is defined [here](https://github.com/canonical/data-platform-libs/blob/3f4aa1ad68b91b1011b4edbb2f0f0397a8b0b188/tests/v0/integration/s3-charm/charmcraft.yaml#L5), which stays untouched.

tests/v0/integration/test_charm.py
reneradoi: Yes, these assertions where useless. The actual check happens afterwards, in the assertion of the relation data to be present (or not). Asserting log messages is error prone and should be avoided in general, I believe. Especially if there are random wait times (3s) for a message to be present. In these cases, it caused test flakeyness and CI instability, and as it doesn't bring value, I decided to remove.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1586,11 +1591,42 @@ def _fetch_relation_data_without_secrets(
...
+            remote_model_uuid = ""
+
+        # if not a cross-model relation we can return data as-is
+        if remote_model_uuid == "" or self._model.uuid == remote_model_uuid:
+            return relation_data
+
+        encryption_key = None
+        if encryption_secret := relation.data[relation.app].get("encryption-secret"):
+            secret = self._model.get_secret(id=encryption_secret)
+            encryption_key = secret.get_content().get("encryption-key")
```
imanenami: nit: we can DRY this out as a private property as it's repeated a couple of times

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -2389,6 +2470,38 @@ def _validate_entity_consistency(event: RelationEvent, diff: Diff) -> None:
                 raise ValueError(f"Cannot change {key} after relation has already been created")
 
     # Event handlers
+    def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
+        """Event emitted when a relation is created."""
+        if not self.relation_data.local_unit.is_leader():
+            return
+
+        try:
+            remote_model = event.relation.remote_model.uuid
```
imanenami: nit: Think this one's also better be called `remote_model_uuid`

imanenami (approved):
Thanks for the effort, looks good to me 👍

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Good idea!

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: True, I've missed a few instances, have changed them all now.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -453,9 +454,9 @@ def _on_subject_requested(self, event: SubjectRequestedEvent):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 58
+LIBPATCH = 59
 
-PYDEPS = ["ops>=2.0.0"]
+PYDEPS = ["ops>=2.0.0", "cryptography"]
```
dragomirp: I would prefer if `cryptography` is inline import with a runtime error instead of a hard dependency. It's a pretty big library a lot of products don't necessarily need and latest versions already dropped support for Python versions on some of our supported platforms (focal with 3.8).

reneradoi: I've removed it.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -433,6 +433,7 @@ def _on_subject_requested(self, event: SubjectRequestedEvent):
     overload,
 )
 
+from cryptography.fernet import Fernet, InvalidToken
```
dragomirp:
```suggestion
try:
    from cryptography.fernet import Fernet, InvalidToken
except ImportError:
    Fernet = None
    InvalidToken = Nonee
```
Conditionally import if cryptography is installed

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1586,11 +1600,37 @@ def _fetch_relation_data_without_secrets(
...
+            return relation_data
+
+        if not (encryption_key := self._get_encryption_key(relation)):
+            return relation_data
+
+        # still here means sensitive data needs to be decrypted
+        for key, value in relation_data.items():
+            if key in CROSS_MODEL_RELATION_CONSUMER_SECRETS:
+                try:
+                    f = Fernet(encryption_key)
```
dragomirp, line 1637: Shouldn't the key be set up outside of the relation items loop?

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1586,11 +1600,37 @@ def _fetch_relation_data_without_secrets(
...
+            logger.warning("Access to remote model failed: %s", e)
+            remote_model_uuid = ""
+
+        # if not a cross-model relation we can return data as-is
+        if remote_model_uuid == "" or self._model.uuid == remote_model_uuid:
+            return relation_data
+
+        if not (encryption_key := self._get_encryption_key(relation)):
+            return relation_data
+
```
dragomirp, line 1626:
```suggestion

        if not Fernet:
            logger.warning("Cryptography module not installed. Could not decrypt sensitive field in cross-model relation")
            return relation_data

```
If there's a key we check if cryptography is available. Could be an exception. Alternatively we can do inline import here and handle or raise the `ImportError`.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1637,8 +1677,30 @@ def _update_relation_data_without_secrets(
...
+        try:
+            remote_model_uuid = relation.remote_model.uuid
+        except (FileNotFoundError, ModelError, RuntimeError):
+            # access to remote model added in Juju 3.6.2, fails with 2.9
+            remote_model_uuid = ""
+
+        if encryption_key and remote_model_uuid != "" and self._model.uuid != remote_model_uuid:
+            for key, value in data.items():
+                try:
+                    f = Fernet(encryption_key)
```
dragomirp: Same as line 1627.

lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -1637,8 +1677,30 @@ def _update_relation_data_without_secrets(
...
             return
 
-        if relation:
-            relation.data[component].update(data)
+        if not relation:
+            return
+
+        # ensure no sensitive information is stored in relation data
+        encryption_key = self._get_encryption_key(relation)
+
```
dragomirp, line 1700:
```suggestion

        if encryption_key and not Fernet:
            logger.warning("Cryptography module not installed. Could not decrypt sensitive field in cross-model relation")
            return relation_data

```
If there's a key we check if cryptography is available. Could be an exception. Alternatively we can do inline import here and handle or raise the `ImportError`.

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Thank you for the suggestion, this should make the change less disruptive for those who don't want to add `cryptography`.

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Recommendation is to keep memory of the Fernet object as low as possible ([it keeps its data in memory](https://cryptography.io/en/latest/fernet/#limitations)). Furthermore I want to keep the error handling block concise and still ensure that, if anything fails, no data is leaked (not relevant here but for updating relation data).

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: Recommendation is to keep memory of the Fernet object as low as possible ([it keeps its data in memory](https://cryptography.io/en/latest/fernet/#limitations)). Furthermore I want to keep the error handling block concise and still ensure that, if anything fails, no data is leaked (not relevant here but for updating relation data).

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: I've taken over your suggestion.

lib/charms/data_platform_libs/v0/data_interfaces.py
reneradoi: I've adjusted the handling a bit here, to ensure no sensitive data is leaked in cases of a cross-model relation but with no cryptography installation.

Mehdi-Bendriss (approved):

dragomirp (approved):
Deployed with [PGB without cryptography](https://github.com/canonical/pgbouncer-operator/pull/661) as a test. LGTM.

PR #273, opened 2026-09-22
fix(DPE-11182): writing None or "" deletes fields

### LINKED ISSUE #272: `write_field` ignores empty values
## Steps to reproduce

The bug is in `OpsRepository.write_field` and `OpsRepository.write_fields`. 

I reproduced the problem with a peer model in a Scenario test:

1. Declare a model with a string field that defaults to empty:
   ```python
   class M(PeerModel):
       name: str = Field(default="")
   ```
2. Build the repository for the peer relation and write a value:
   ```python
   repo = OpsPeerRepositoryInterface(model, "peers", M).repository(rel_id, app)
   m = build_model(repo, M)
   m.name = "x"
   write_model(repo, m)
   ```
3. Read the databag. `name` is `"x"`.
4. Clear the field and write again:
   ```python
   m.name = ""
   write_model(repo, m)
   ```
5. Read the databag. `name` is still `"x"`.

Second reproduction, for `write_fields`:

1. Call `repo.write_fields({"a": "1", "b": "2"})`.
2. Read the databag. Neither key exists.

## Expected behavior

In step 5 the `name` key is removed from the databag, or at least the caller gets an error or a warning. Juju removes a key when it is set to an empty string (`relation-set name=""`), and ops does the same in `RelationDataContent._update_cache`:

```python
if value == '':
    # Match the behavior of Juju, which is that setting the value to an
    # empty string will remove the key entirely from the relation data.
    self._data.pop(key, None)
```

`write_fields` step 2: both keys are written.

## Actual behavior

`write_field` returns before touching the databag when the value is falsy:

```python
def write_field(self, field: str, value: Any) -> None:
    if not self.relation:
        logger.info("No relation to get value from")
        return
    if self.component not in self.relation.data:
        logger.info(f"Component {self.component} not in relation {self.relation}")
        return
    if not value:
        return
    self.relation.data[self.component].update({field: value})
```

The two earlier early returns log a message. The `if not value` return does not, and the abstract method has no docstring that mentions the rule. `write_model` deletes a key only when the dumped value is `None`, so a field with an empty-string default can be set once and never cleared. Any custom serializer that produces an empty string, for example a `set[str]` joined with commas, has the same problem.

`write_fields` is a generator expression that nothing iterates, so it never calls `write_field`:

```python
def write_fields(self, mapping: dict[str, Any]) -> None:
    ...
    (self.write_field(field, value) for field, value in mapping.items())
```

## Versions

Operating system: `Ubuntu 26.04.1 LTS`

Juju CLI: `3.6.28-genericlinux-amd64`

Juju agent: not needed; reproduces in unit tests

Library versions: dpcharmlibs-interfaces 1.1.0, data_interfaces v1 LIBPATCH 3, ops 3.8.1

## Log output

Juju debug log: not applicable, there is no log line for the skipped write. That is part of the bug.

## Additional context

I noticed this while revewing the OpenSearch charm migrating from data_interfaces v0 to v1. The v0 `Data.update_relation_data` path treated a falsy value as a delete, so call sites written against v0 keep stale values on v1 with no error. 

I think we should either:
1. Delete the key when the value is an empty string, to match Juju and ops.
2. Raise `ValueError` on an empty value.
3. Keep the skip, but log at `WARNING` and document on `AbstractRepository.write_field` and in the `write_model` docstring that `None` is the only way to delete.

I am leaning towards 1 if it doesn't break any charms because I believe most charm users would expect that based on juju and ops behavior.

skourta:
fixes #272.

`OpsRepository.write_field` returned early on a falsy value, so a field that was set once could never be cleared through `write_field` or `write_model`.
Juju, ops and the v0 lib all remove a key that is set to an empty string, so charms ported from v0 kept stale values with no error or log line.
`write_fields` built a generator that nothing iterated, so it never wrote anything. `delete_fields` had the same bug and is fixed here too.

## Changes

- `write_field`: `None` or `""` now removes the key through `delete_field`,   with a debug log. Other values are passed to ops as before.
- `write_fields` and `delete_fields`: replaced the generator expressions with `for` loops.
- The v0 compatibility write-back in `ResourceProviderEventHandler` is now skipped when the requirer has not sent a resource alias yet. The old falsy check hid a `write_field("", "")` call in that case.
- Documented the empty-value behaviour on `AbstractRepository.write_field`, `write_fields` and `write_model`.
- `LIBPATCH` 3 -> 4.

## Tests

Four new unit tests in `tests/v1/unit/test_data_interfaces.py`, run for both the app and unit peer scopes: empty `write_field` removes the key, `write_fields` writes every key, `delete_fields` removes every key, and the `write_model` reproduction from the issue.


dragomirp:
Shouldn't the PR go to [data-platform-charmlibs](https://github.com/canonical/data-platform-charmlibs) repo instead?

skourta:
> Shouldn't the PR go to [data-platform-charmlibs](https://github.com/canonical/data-platform-charmlibs) repo instead?

Sorry my bad. I opened one [there](https://github.com/canonical/data-platform-charmlibs/pull/29).

